PR #7, opened 2022-12-14
Remove code that prepares the local nuget package for tests

Zaid-Ajaj:
This removes the function `preparePulumiSdkNugetLocally()` which was to build a local nuget package of the Pulumi SDK for testing inside `~/.pulumi-dev/nuget`. This is no longer needed because the integration tests now use a local project reference instead

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,8 +69,8 @@ And finally, `pulumi preview` and `pulumi up` as you would any other Pulumi proj
 
 ### Prerequisites
 - Dotnet [SDK v6+](https://dotnet.microsoft.com/download/dotnet/6.0) installed on your machine.
-- Go 1.19 or later (for building and testing `pulumi-language-dotnet`)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -69,8 +69,8 @@ And finally, `pulumi preview` and `pulumi up` as you would any other Pulumi proj
 
 ### Prerequisites
 - Dotnet [SDK v6+](https://dotnet.microsoft.com/download/dotnet/6.0) installed on your machine.
-- Go 1.19 or later (for building and testing `pulumi-language-dotnet`)
```
Frassle, line 72: These were informative, why not keep?

Frassle (approved):

README.md
Zaid-Ajaj: You are right, I've brought them back 😄

PR #8, opened 2022-12-14
Use local language plugin for integration tests

Zaid-Ajaj:
Implements `installLanguagePluginLocally()` function which builds `pulumi-language-dotnet` plugin and copies it to `~/.pulumi-dev/bin`. Then prepends the `PATH` with `~/.pulumi-dev/bin` during the process run. 

Used implicitly by:
 - dotnet run integration test <testName> (runs in CI)
 - dotnet run all-integration-tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,9 @@ dotnet run build-language-plugin
 # Testing the language plugin
 dotnet run test-language-plugin
 
+# Install the language plugin into ~/.pulumi-dev/bin and prepend it to your PATH (for testing)
+dotnet run install-language-plugin
+
 # Sync proto files from pulumi/pulumi
 dotnet run sync-proto-files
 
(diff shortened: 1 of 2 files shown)


Frassle:
I need to have a play but I think we can use plugin link to do this instead

Frassle:
OK plugin link is not quite connected in well enough (especially to program test) to do this, sad.
However I think I'd rather look at us using the binary in the repo via PATH rather than asking the user to install the binary and set PATH themselves.
Testing should not require manipulation of state external to the repo (ideally), and it should _just work_ for new users as much as possible.
I think if we tweak the tests to call a dotnetProgramTest rather than ProgramTest directly and have that:
A) Set PrepareProject to prepareDotnetProject
B) Set Env["PATH"] to "./pulumi-language-dotnet:$PATH"
That should work and not require the user to know about anything extra.

PR #9, opened 2022-12-14
Add test that ensures invoking `pulumi:pulumi:getResource` keeps resource refs

justinvp:
Ported from https://github.com/pulumi/pulumi/pull/11522

Following up from https://github.com/pulumi/pulumi/pull/11382

Part of https://github.com/pulumi/pulumi/issues/11384


Zaid-Ajaj (approved):
LGTM! Can you please add the test to the list of integration tests in `.github/workflows/pr.yml`? This way it runs on CI as well 😄 

I will replace this list later with `dotnet run all-integration-tests` probably, I did this because it gives a nice view of which tests failed when they do but keeping it in sync might be a hassle

justinvp:
> Can you please add the test to the list of integration tests in `.github/workflows/pr.yml`?

Done

> it gives a nice view of which tests failed

That is nice, though, I wish they'd run in parallel rather than one at a time.

Zaid-Ajaj:
> I wish they'd run in parallel rather than one at a time.

@justinvp I tried this at first but when running multiple tests that use the same local reference to the SDK, it becomes really flaky and fails at random 😞 this was my compromise to make them reliable given that there aren't that many integration tests

PR #10, opened 2022-12-15
Update Pulumi projects to net6.0

Frassle:
.NET Core 3.1 is out of support on December 13th:
https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core

We've already stopped testing on .NET Core 3.1, and program gen has been targeting only 6.0 for a while now as well.

This updates the core porjects to only 6.0 as well.

We'll still want to do https://github.com/pulumi/pulumi/pull/11543 in pu/pu to update sdkgen.

### DIFF
--- a/sdk/Pulumi/Core/Output.cs
+++ b/sdk/Pulumi/Core/Output.cs
@@ -3,7 +3,6 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
-using System.Diagnostics;
 using System.Linq;
 using System.Text.Json;
 using System.Text.Json.Serialization;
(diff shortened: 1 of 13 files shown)


Frassle:
hah null reference checks are now running, got some annotations to add it seems

sdk/Pulumi/Core/Output.cs
```diff
@@ -285,7 +267,8 @@ public sealed class Output<T> : IOutput
 
             if (Deployment.TryGetInternalInstance(out var instance))
             {
-                instance.Runner.RegisterTask(TypeNameHelper.GetTypeDisplayName(GetType(), false), dataTask);
+                var description = GetType().ToString();
```
Frassle, line 270: `TypeNameHelper` doesn't exist in 6.0 (Looks like it got made internal)

Zaid-Ajaj (approved):

PR #76, opened 2023-01-06
Provider support for dotnet

Frassle:
This adds a new namespace to the dotnet sdk "Experimental.Provider" with classes to support the writing of a native dotnet provider.

The Provider class will internally deal with our grpc protocol and call into virtual methods given by the user which use only dotnet native domain models.

I've changed the "testprovider" in the integration tests to be written using this framework, rather than being written in Go. I've also added support for the dotnet language host to support RunPlugin so shimless works against that test provider.
All the current integration tests pass, so the framework covers at least those basics.

It is a bit sad all this is manually written, I feel like there _might_ be a good way to codegen this from the protobuf specs such that if they ever change this file could stay in sync.

Also the PropertyValue Marshal/Unmarshal duplicates a lot of the existing serialisation logic, and I think that logic could be rewritten as a layer ontop of this (that is we'd have "objects -> PropertyValues" and "PropetyValues -> grpc" rather than what we currently have which is "objects -> grpc" and now "PropertyValues -> grpc")


Zaid-Ajaj (approved):
This looks really great, very excited to try it out ✅ 

One question: does this being experimental mean we can add semi-breaking changes in a separate PR? Because looking at the provider implementation, wondering if we pass it two things: 
- `ILogger` from Microsoft.Extensions.Logging
- `IConfiguration` from Microsoft.Extensions.Configuration

Both of which are very useful and play well with the dotnet ecosystem (i.e. you can use Serilog for logging and pass that as `ILogger` for the provider). 

Though these aren't necessarily needed at the moment

Frassle:
> One question: does this being experimental mean we can add semi-breaking changes in a separate PR?

Yes, we can make breaking changes to this later. I'll add a changelog for this and explain that in the changelog, probably write a quick blog post for this as well.

Intention is this is a first look preview for the community to try and comment on. Well work those comments in then mark this stable.

PR #78, opened 2023-01-16
Add a test for the DeletedWith resource option

justinvp:
Add a test for the `DeletedWith` resource option to ensure it works as expected.

Also fixed a bug in the Provider's `Delete`.


sdk/Pulumi/Provider/Provider.cs
```diff
@@ -863,7 +863,7 @@ public override async Task<Empty> Delete(Pulumirpc.DeleteRequest request, Server
         {
             try
             {
-                var domRequest = new DeleteRequest(request.Urn, request.Id, PropertyValue.Marshal(request.Properties), TimeSpan.FromSeconds(request.Timeout));
+                var domRequest = new DeleteRequest(request.Id, request.Urn, PropertyValue.Marshal(request.Properties), TimeSpan.FromSeconds(request.Timeout));
```
justinvp, line 866: We were passing the wrong thing to the `DeleteRequest` ctor. I noticed that we have a handful of `Request` classes that take `urn` and `id` in the ctor, but their order isn't consistent across the classes. Some take `urn` first, others take `id` first. Should we change them to have a consistent order?

Frassle: Yes URN first probably.

integration_tests/integration_dotnet_test.go
```diff
@@ -486,8 +486,8 @@ func TestProvider(t *testing.T) {
 	testDotnetProgram(t, &integration.ProgramTestOptions{
 		Dir:            filepath.Join("provider"),
 		LocalProviders: []integration.LocalDependency{{Package: "testprovider", Path: "testprovider"}},
-		Verbose: true,
-		DebugLogLevel: 10,
+		Verbose:        true,
```
Frassle, line 489: Oh these were from debugging, Verbose and DebugLogLevel can be removed from this test.

.github/workflows/pr.yml
```diff
@@ -103,4 +103,6 @@ jobs:
       - name: TestDotNetAliasesRenameComponent
         run: dotnet run integration test TestDotNetAliasesRenameComponent
       - name: TestDotNetAliasesRetypeParents
-        run: dotnet run integration test TestDotNetAliasesRetypeParents
\ No newline at end of file
+        run: dotnet run integration test TestDotNetAliasesRetypeParents
+      - name: TestDeletedWith
```
Frassle, line 107: Oh can we add "TestProvider" here as well, I missed that in the last change.

cc @Zaid-Ajaj is there a reason these are explicitly listed out rather than just running every integration test?

Zaid-Ajaj: @Frassle just for better visibility on each integration that gets shown in the github action. I think we can swap these out for `dotnet run all-integration-tests`

sdk/Pulumi/Provider/Provider.cs
Frassle: Fixed up in https://github.com/pulumi/pulumi-dotnet/pull/83

integration_tests/integration_dotnet_test.go
Frassle: Fixed up in https://github.com/pulumi/pulumi-dotnet/pull/82

Zaid-Ajaj (approved):
LGTM

PR #80, opened 2023-01-18
[sdk] Delegates alias computation to engine

### LINKED ISSUE #14: [sdks/dotnet] Delegate alias computation to the engine
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Currently, we pre-compute all possible alias combinations in the dotnet sdk [here](https://github.com/pulumi/pulumi/blob/09d67511dd356fa7b77c7d0ac03d40b99d22c937/sdk/dotnet/Pulumi/Resources/Resource.cs#L341). When each resource in an inheritance chain has many aliases, this quickly grows multiplicatively. Instead, we should simply compute the aliases for the resource we're registering, and allow the engine to do this computation for us. Additionally, we should ensure we send the structured version of aliases across rather than just raw URNs.

### Affected area/feature

Dotnet SDK

Zaid-Ajaj:
Fixes #14 

Removes `AllAliases` function and instead constructs a`List<Pulumirpc.Alias>` in `PrepareAliases` (linear time) which populates the `RegisterResourceRequest` accordingly:
```csharp
if (prepareResult.SupportsAliasSpec) 
{
    request.Aliases.AddRange(prepareResult.Aliases);
}
else 
{
    var aliasUrns = prepareResult.Aliases.Select(a => a.Urn);
    request.AliasURNs.AddRange(aliasUrns);
}
```
Wasn't exactly sure how to test this other than constructing a lot (=1000) of aliases in `AliasesMocks` and _observe_ that it doesn't just hang for a while. 

All integration tests are succeeding locally except for `TestProvider` but that seems unrelated. Let's see what CI has to say about it.

### DIFF
--- a/sdk/Pulumi/Deployment/Deployment_Prepare.cs
+++ b/sdk/Pulumi/Deployment/Deployment_Prepare.cs
@@ -1,8 +1,9 @@
 // Copyright 2016-2021, Pulumi Corporation
 
+using System;
+using System.Linq;
 using System.Collections.Generic;
 using System.Collections.Immutable;
-using System.Linq;
 using System.Threading.Tasks;
 using Google.Protobuf.WellKnownTypes;
 
@@ -115,20 +116,8 @@ namespace Pulumi
                 propertyToDirectDependencyUrns[propertyName] = urns;
             }
 
-            // Wait for all aliases. Note that we use 'res._aliases' instead of 'options.aliases' as
-            // the former has been processed in the Resource constructor prior to calling
-            // 'registerResource' - both adding new inherited aliases and simplifying aliases down
-            // to URNs.
-            var aliases = new List<string>();
-            var uniqueAliases = new HashSet<string>();
-            foreach (var alias in res._aliases)
-            {
-                var aliasVal = await alias.ToOutput().GetValueAsync(whenUnknown: "").ConfigureAwait(false);
-                if (aliasVal != "" && uniqueAliases.Add(aliasVal))
-                {
-                    aliases.Add(aliasVal);
-                }
-            }
+            var resourceMonitorSupportsAliasSpec = await MonitorSupportsAliasSpec().ConfigureAwait(false);
+            var aliases = await PrepareAliases(res, options, resourceMonitorSupportsAliasSpec).ConfigureAwait(false);
 
             return new PrepareResult(
                 serializedProps,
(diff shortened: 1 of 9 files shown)


Frassle:
Run `dotnet format` :) Also we should split that to it's own job so the unit tests can still run and give signal even if formatting is wrong.

sdk/Pulumi/Deployment/Deployment_Prepare.cs
```diff
@@ -146,6 +136,183 @@ void LogExcessive(string message)
...
+                        // alias contains unknowns, skip it.
+                        continue;
+                    }
+
+                    var aliasName = await Resolve(resolvedAlias.Name, whenUnknown: "");
+                    var aliasType = await Resolve(resolvedAlias.Type, whenUnknown: "");
+                    var aliasStack = await Resolve(resolvedAlias.Stack, Instance.StackName);
+                    var aliasProject = await Resolve(resolvedAlias.Project, Instance.ProjectName);
+
+                    if (string.IsNullOrEmpty(aliasName) || string.IsNullOrEmpty(aliasType))
```
Frassle, line 169: This doesn't seem right. Quite often you'll give an alias with just a name and so type will be ""

Zaid-Ajaj: In `CollapseAliasToUrn` we have
```csharp
var name = a.Name ?? defaultName;
var type = a.Type ?? defaultType;

// later... 

if (name == null)
    throw new ArgumentNullException("No valid 'Name' passed in for alias.");
if (type == null)
    throw new ArgumentNullException("No valid 'type' passed in for alias.");
```
Which suggested that both `name` _and_ `type` must be non-null so I wasn't sure. I will change it to:
```csharp
var aliasSpec = new Pulumirpc.Alias.Types.Spec
{
    Name = await Resolve(resolvedAlias.Name, resource.GetResourceName()),
    Type = await Resolve(resolvedAlias.Type, resource.GetResourceType()),
    Stack = await Resolve(resolvedAlias.Stack, Instance.StackName),
    Project = await Resolve(resolvedAlias.Project, Instance.ProjectName),
};
```

Zaid-Ajaj:
@Frassle 
 - Using `dotnet run format-sdk` changes every file I have, removing a whitespace at the end 😓 
 - Integration test `TestProvider` is failing on the `destroy` step.

Frassle:
> Using dotnet run format-sdk changes every file I have, removing a whitespace at the end 😓

Odd? Shouldn't this be consistent across machines?

> Integration test TestProvider is failing on the destroy step.

OK leave the test off the test list for now, I'll get that fixed.

sdk/Pulumi/Deployment/Deployment_Prepare.cs
Frassle: I'm pretty sure the engine is happy to take an Alias object with type or name as "".

Zaid-Ajaj:
> Odd? Shouldn't this be consistent across machines?

@Frassle I've added a minimal .editorconfig which is now formatting the same locally and on CI

> OK leave the test off the test list for now, I'll get that fixed.

Removed that test the time being

> I'm pretty sure the engine is happy to take an Alias object with type or name as "".

Alright, then using empty string as default, see last commit:
```csharp
var aliasSpec = new Pulumirpc.Alias.Types.Spec
{
    Name = await Resolve(resolvedAlias.Name, ""),
    Type = await Resolve(resolvedAlias.Type, ""),
    Stack = await Resolve(resolvedAlias.Stack, Instance.StackName),
    Project = await Resolve(resolvedAlias.Project, Instance.ProjectName),
};
```

.editorconfig
```diff
@@ -0,0 +1,3 @@
+# C# files
```
Frassle, line 1: Can you change this in it's own PR, it's pretty much ruined the diff for this PR.

Zaid-Ajaj: OK let me try, because before adding the file I also got tons of diffs when I ran `dotnet run format-sdk`

Zaid-Ajaj: That did not help at all 😞 now more diffs. Maybe `dotnet format` is not the same across machines _by default_ and it does need config?

Frassle: Yes do the format change in another PR, I'll approve that then rebase this

sdk/Pulumi/Deployment/Deployment_Prepare.cs
```diff
@@ -146,6 +136,173 @@ void LogExcessive(string message)
...
+                    {
+                        // alias contains unknowns, skip it.
+                        continue;
+                    }
+
+                    var aliasSpec = new Pulumirpc.Alias.Types.Spec
+                    {
+                        Name = await Resolve(resolvedAlias.Name, ""),
+                        Type = await Resolve(resolvedAlias.Type, ""),
+                        Stack = await Resolve(resolvedAlias.Stack, Instance.StackName),
```
Frassle, line 168: These can actually default to "" as well, the engine will fill them in.

Zaid-Ajaj: done ✅

Frassle (approved):

PR #82, opened 2023-01-19
Work around old engines port parsing bug

Frassle:
Also make sure we're running integration tests on Windows and MacOS not just linux, and add the Provider test to the test list.

### DIFF
--- a/integration_tests/integration_dotnet_test.go
+++ b/integration_tests/integration_dotnet_test.go
@@ -326,7 +326,6 @@ func TestConstructPlainDotnet(t *testing.T) {
 	expectedResourceCount := 8
 
 	localProviders := []integration.LocalDependency{
-		{Package: "testprovider", Path: "testprovider"},
 		{Package: "testcomponent", Path: filepath.Join(testDir, componentDir)},
 	}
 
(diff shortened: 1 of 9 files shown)


integration_tests/integration_dotnet_test.go
```diff
@@ -326,7 +326,6 @@ func TestConstructPlainDotnet(t *testing.T) {
 	expectedResourceCount := 8
 
 	localProviders := []integration.LocalDependency{
-		{Package: "testprovider", Path: "testprovider"},
```
Frassle, line 329: Seems this wasn't actually used.

integration_tests/integration_dotnet_test.go
```diff
@@ -486,8 +485,6 @@ func TestProvider(t *testing.T) {
 	testDotnetProgram(t, &integration.ProgramTestOptions{
 		Dir:            filepath.Join("provider"),
 		LocalProviders: []integration.LocalDependency{{Package: "testprovider", Path: "testprovider"}},
-		Verbose: true,
```
Frassle, line 489: Left overs from setting this test up.

Zaid-Ajaj (approved):
lgtm

PR #84, opened 2023-01-19
Consistently case Id in provider interface

Frassle:
Other places in the Pulumi SDK use "Id" not "ID". That is also inline with Microsoft naming guidance.


Frassle:
Should rebase on top of https://github.com/pulumi/pulumi-dotnet/pull/83 once that merges

Zaid-Ajaj (approved):

PR #88, opened 2023-01-23
[sdk] Correctly check for alias support in the engine and map fully specified alias urns

Zaid-Ajaj:
Learnings from implementing go alias delegation to the engine:
 - the flag is called "aliasSpecs" not "aliasSpec"
 - aliases can also provide a fully qualified URN which can be passed as is to the engine

### DIFF
--- a/sdk/Pulumi/Deployment/Deployment.cs
+++ b/sdk/Pulumi/Deployment/Deployment.cs
@@ -226,9 +226,9 @@ namespace Pulumi
         /// Returns whether the resource monitor we are connected to supports the "aliasSpec" feature across the RPC interface.
         /// In which case we no longer compute alias combinations ourselves but instead delegate the work to the engine.
         /// </summary>
-        internal Task<bool> MonitorSupportsAliasSpec()
+        internal Task<bool> MonitorSupportsAliasSpecs()
         {
-            return MonitorSupportsFeature("aliasSpec");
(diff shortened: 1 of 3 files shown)


sdk/Pulumi/Deployment/Deployment.cs
```diff
@@ -226,9 +226,9 @@ internal Task<bool> MonitorSupportsDeletedWith()
         /// Returns whether the resource monitor we are connected to supports the "aliasSpec" feature across the RPC interface.
         /// In which case we no longer compute alias combinations ourselves but instead delegate the work to the engine.
         /// </summary>
-        internal Task<bool> MonitorSupportsAliasSpec()
+        internal Task<bool> MonitorSupportsAliasSpecs()
         {
-            return MonitorSupportsFeature("aliasSpec");
+            return MonitorSupportsFeature("aliasSpecs");
```
Frassle, line 231: oh well this is sad
I'll raise a PR for the protobuf to say we should stop adding more strings here and generate an enum for engine features instead. I don't think we want to try moving the existing flags, but if we add anymore it'll be good to see a note reminding us about this.

Frassle (approved):

sdk/Pulumi/Deployment/Deployment.cs
Zaid-Ajaj: An enum would be great for this

PR #101, opened 2023-02-03
feature: Add outputs yielding sp test stack func

dpwdec:


Frassle:
A method can't overload based on just a change in return type. This will need to be renamed.

PR #109, opened 2023-02-27
sdk: Test case for Provider propagation

abhinav:
Adds a test case to verify that a ComponentResource
propagates its Provider and Providers values
to children constructed with that type.

Refs pulumi/pulumi#12161

### DIFF
--- a/sdk/Pulumi.Tests/Resources/ComponentResourceTests.cs
+++ b/sdk/Pulumi.Tests/Resources/ComponentResourceTests.cs
@@ -1,0 +1,116 @@
+// Copyright 2016-2023, Pulumi Corporation
+
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Pulumi.Testing;
+using Xunit;
+
+namespace Pulumi.Tests.Resources
+{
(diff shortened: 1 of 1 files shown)


sdk/Pulumi.Tests/Resources/ComponentResourceTests.cs
```diff
@@ -0,0 +1,104 @@
...
+                    "custom",
+                    ResourceArgs.Empty,
+                    new CustomResourceOptions
+                    {
+                        Parent = component,
+                    }
+                );
+
+                Assert.Equal(provider, custom._provider);
+            });
```
dixler, line 47: Could you add a comment to clarify the motivation of what's going on here. It makes sense at the end when you assert equal, but it's not immediately clear and you have to keep a lot of stuff in your head if you were reading it in the codebase without the PR context.

abhinav: Good point. I realize that the context from the commit/issue wasn't carried over to the code. 
Will fix.

justinvp (approved):

PR #113, opened 2023-03-02
Update setup-dotnet and checkout

Frassle:
These were warning about still being Node.js 12 actions which are deprecated.


abhinav:
FYI dependabot supports auto-updating actions. 
I'll follow up with a PR for that.

PR #119, opened 2023-03-17
Fix automation tests for forks

Frassle:


sdk/Pulumi.Automation.Tests/CommandExceptionTests.cs
```diff
@@ -9,8 +10,33 @@
 
 namespace Pulumi.Automation.Tests
 {
-    public class CommandExceptionTests
+    [Collection("PULUMI_BACKEND_URL")]
```
Frassle, line 13: Can't run these tests in parallel with any other tests using PULUMI_BACKEND_URL

sdk/Pulumi.Automation.Tests/LocalWorkspaceTests.cs
```diff
@@ -2098,7 +2122,8 @@ private class StackStateResource
             public bool Protect { get; set; }
         }
 
-        [Fact]
+        // This test only works with the Pulumi Service backend.
+        [ServiceApiFact]
```
Frassle, line 2126: Tags are a service feature

Issue #126, opened 2023-04-05
Pulumi Automation API throws Grpc.Core.RpcException


klyse:
### What happened?

I have a ASP.net core 7 project that uses the inline Pulumi automation API.

Example:

```cs
var stack = await GetStackAsync(vm, cancellationToken);
await stack.CancelAsync(token);
await stack.RefreshAsync(cancellationToken: token);
await stack.UpAsync(cancellationToken: token);
```

I use Sentry as Error tracking system. I'm getting a log of errors (16k in the last 24h) that I only partially understand. They seem benign.

```
Grpc.Core.RpcException: Status(StatusCode="Unavailable", Detail="Error starting gRPC call. HttpRequestException: Connection refused (127.0.0.1:37649) SocketException: Connection refused", DebugException="System.Net.Http.HttpRequestException: Connection refused (127.0.0.1:37649)
 ---> System.Net.Sockets.SocketException (111): Connection refused
   at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.ThrowException(SocketError error, CancellationToken cancellationToken)
   at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.System.Threading.Tasks.Sources.IValueTaskSource.GetResult(Int16 token)
   at System.Net.Sockets.Socket.<ConnectAsync>g__WaitForConnectWithCancellation|281_0(AwaitableSocketAsyncEventArgs saea, ValueTask connectTask, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.ConnectToTcpHostAsync(String host, Int32 port, HttpRequestMessage initialRequest, Boolean async, CancellationToken cancellationToken)
   --- End of inner exception stack trace ---
   at System.Net.Http.HttpConnectionPool.ConnectToTcpHostAsync(String host, Int32 port, HttpRequestMessage initialRequest, Boolean async, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.ConnectAsync(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.AddHttp2ConnectionAsync(QueueItem queueItem)
   at System.Threading.Tasks.TaskCompletionSourceWithCancellation`1.WaitWithCancellationAsync(CancellationToken cancellationToken)
[...]

PR #127, opened 2023-04-06
Wait for async work to finish even in the error case

Frassle:
Should help with https://github.com/pulumi/pulumi-dotnet/issues/126 but might not be a full fix.


sdk/Pulumi/Deployment/Deployment.Runner.cs
```diff
@@ -191,7 +217,6 @@ private async Task<bool> LogExceptionToErrorStream(Exception exception)
                 {
                     // We got an error while logging itself. Nothing
                     // to do here but print some errors and abort.
-                    _deploymentLogger.LogError(exception, "Error occurred trying to send logging message to engine");
```
Frassle, line 194: No point trying to log if we're hitting errors trying to log

PR #159, opened 2023-07-13
add a global.json to specifiy a min sdk

SimonCropp:


sdk/global.json
```diff
@@ -0,0 +1,7 @@
+﻿{
+  "sdk": {
+    "version": "7.0.306",
```
Frassle, line 3: We're currently targeting dotnet 6, and that's what CI is running. Can this be set to "6.0.412"?

SimonCropp: you should be able to target net 6 using the net7 sdk

Frassle: We could but we'd need [CI updated to 7](https://github.com/pulumi/pulumi-dotnet/blob/main/.github/workflows/pr.yml#L28-L31) and I think I'd rather test against exactly the minimum version we say we support. We should probably be testing against the latest as well (which we're currently missing), but I'd assume this would be fine if global.json asked for 6 but only the 7 sdk was installed.

PR #165, opened 2023-08-06
[sdk] Implement Converter abstraction for building converter plugins

Zaid-Ajaj:
This PR implements a `Converter` abstraction under the `Pulumi.Experimental.Converter` namespace which allows easily building and serving converter plugins from dotnet.

Minimal example:
```cs
using Pulumi.Experimental.Converter;

var converter = Converter.CreateSimple(request =>
{
    return ConvertProgramResponse.Empty;
});

await Converter.Serve(converter);
```
In F#:
```fs
open Pulumi.Experimental.Converter

let converter = Converter.CreateSimple (fun request -> ConvertProgramResponse.Empty)

converter
|> Converter.Serve
|> Async.AwaitTask
|> Async.RunSynchronously
```
Here, you give the converter implementation to `Converter.CreateSimple`, this gives you an instance of `SimpleProgramConverter` which implements `Converter`. Then you use that instance as an argument for `Converter.Serve`

Additionally, users can inherit `Converter` in a dedicated class and override the `ConvertProgram` function. 

The `ConvertState` function is not added yet, that can be done in a separate PR when necessary. 

Note that I executed `dotnet run sync-proto-files` to get latest additions from Pulumi (including the converter definitions)

### DIFF
--- a/sdk/Pulumi/Converter/Converter.cs
+++ b/sdk/Pulumi/Converter/Converter.cs
@@ -64,7 +64,6 @@ namespace Pulumi.Experimental.Converter
         }
 
         public static async Task Serve(
-            string[] args,
             Converter converter,
             CancellationToken? cancellationToken = null,
             TextWriter? stdout = null)
@@ -86,14 +85,6 @@ namespace Pulumi.Experimental.Converter
                             // clear so we don't read appsettings.json
                             // note that we also won't read environment variables for config
                             config.Sources.Clear();
-
-                            var memConfig = new Dictionary<string, string>();
-                            if (args.Length > 0)
-                            {
-                                memConfig.Add("Host", args[0]);
-                            }
-
-                            config.AddInMemoryCollection(memConfig);
                         })
                         .ConfigureLogging(loggingBuilder =>
                         {
--- a/sdk/Pulumi/Converter/Converter.cs
+++ b/sdk/Pulumi/Converter/Converter.cs
@@ -1,0 +1,246 @@
+using System;
+using System.Net;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Diagnostics;
+using System.Linq;
+using System.Collections.Generic;
+using Grpc.Core;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Hosting;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Hosting.Server;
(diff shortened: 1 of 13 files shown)


Frassle:
We should copy the same style as Experimental.Provider here.
That is interface based and with read-only records for requests (e.g. see CheckRequest at https://github.com/pulumi/pulumi-dotnet/blob/main/sdk/Pulumi/Provider/Provider.cs#L24-L46)

sdk/Pulumi/Converter/Converter.cs
```diff
@@ -0,0 +1,259 @@
...
+        public static async Task Serve(
+            string[] args,
+            Converter converter,
+            CancellationToken? cancellationToken = null,
+            TextWriter? stdout = null)
+        {
+            // maxRpcMessageSize raises the gRPC Max message size from `4194304` (4mb) to `419430400` (400mb)
+            var maxRpcMessageSize = 400 * 1024 * 1024;
+
+            var host = Host.CreateDefaultBuilder()
```
Frassle, line 75: We ought to find a way to share this between Converters and Providers. Every plugin is going to look like this in the end.

sdk/Pulumi/Converter/Converter.cs
```diff
@@ -0,0 +1,259 @@
...
+            {
+                var response = convertProgram(request);
+                return Task.FromResult(response);
+            });
+        }
+
+        public static Converter CreateSimpleAsync(Func<ConvertProgramRequest, Task<ConvertProgramResponse>> convertProgramAsync)
+        {
+            return new SimpleProgramConverter(convertProgramAsync);
+        }
```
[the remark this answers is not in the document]
Zaid-Ajaj: If we rewrite all of our {lang}2pulumi converters, they will only need this `CreateSimple` method. Almost zero boilerplate to work with

sdk/Pulumi/Converter/Converter.cs
Zaid-Ajaj: Some parts can be shared, especially where we get the port and print it out. I was wondering though, does the engine send any args to converters plugins? I am thinking this piece of code is redundant:
```cs
var memConfig = new Dictionary<string, string>();
if (args.Length > 0)
{
    memConfig.Add("Host", args[0]);
}
config.AddInMemoryCollection(memConfig);
```
Which could also mean that we can remove the `string[] args` parameter from `Serve`

Frassle: > does the engine send any args to converters plugins?

Not at the moment.

Zaid-Ajaj: Nice, so it can be removed from the API 😄 if we decide that it is needed later on, we can just call `Environment.GetCommandLineArgs()`

Frassle (approved):

Issue #166, opened 2023-08-11
Automation API sometimes has a JSON exception and crashes the process


JoeStead:
### What happened?

When running a deploy, or preview using the Automation API, there seems to be a transient fault that crashes the process.

This is _after_ the preview/up command has completed, so the change is successfully applied but crashes after completion. A re-run tends to fix it.

```
Unhandled exception. System.Text.Json.JsonException: Expected end of string, but instead reached end of data. Path: $ | LineNumber: 0 | BytePositionInLine: 1252.
 ---> System.Text.Json.JsonReaderException: Expected end of string, but instead reached end of data. LineNumber: 0 | BytePositionInLine: 1252.
   at System.Text.Json.ThrowHelper.ThrowJsonReaderException(Utf8JsonReader& json, ExceptionResource resource, Byte nextByte, ReadOnlySpan`1 bytes)
   at System.Text.Json.Utf8JsonReader.ConsumeStringAndValidate(ReadOnlySpan`1 data, Int32 idx)
   at System.Text.Json.Utf8JsonReader.ConsumeString()
   at System.Text.Json.Utf8JsonReader.ConsumeValue(Byte marker)
   at System.Text.Json.Utf8JsonReader.ReadSingleSegment()
   at System.Text.Json.Utf8JsonReader.Read()
   at System.Text.Json.Utf8JsonReader.TrySkip()
   at System.Text.Json.JsonSerializer.Read[TValue](Utf8JsonReader& reader, JsonTypeInfo jsonTypeInfo)
   --- End of inner exception stack trace ---
   at System.Text.Json.ThrowHelper.ReThrowWithPath(ReadStack& state, JsonReaderException ex)
   at System.Text.Json.JsonSerializer.Read[TValue](Utf8JsonReader& reader, JsonTypeInfo jsonTypeInfo)
   at System.Text.Json.JsonSerializer.Deserialize[TValue](Utf8JsonReader& reader, JsonSerializerOptions options)
   at Pulumi.Automation.Serialization.Json.MapToModelJsonConverter`2.Read(Utf8JsonReader& reader, Type typeToConvert, JsonSerializerOptions options)
   at System.Text.Json.Serialization.JsonConverter`1.TryRead(Utf8JsonReader& reader, Type typeToConvert, JsonSerializerOptions options, ReadStack& state, T& value)
[...]

PR #167, opened 2023-08-11
Defend against invalid event data being parsed as json

JoeStead:
Adds some guards in a non-destructive way against invalid event data.

Will hopefully resolve: https://github.com/pulumi/pulumi-dotnet/issues/166


Zaid-Ajaj:
Running GH workflows now

Zaid-Ajaj:
Published as of v3.56.1 🚀 thanks again for the contribution 🙏

PR #183, opened 2023-10-06
[sdk] Register and await tasks created from `Apply` that don't return anything.

### LINKED ISSUE #139: Functions inside an apply do not execute
### What happened?

I am creating an example that provisions a lambda function and then uses the resolved outputs to run a function that deletes old lambda versions.

The code inside the apply gets swallowed and never executes.

### Expected Behavior

I would expect the code inside the apply to run correctly.

### Steps to reproduce

See this example: https://github.com/pierskarsenbarg/lee-pulumi-examples/blob/6cfa63026c41c085aa0860cad21853744bf7b32d/dotnet/aws/lambda_delete_old_versions/infrastructure/MyStack.cs#LL96C3-L96C3

### Output of `pulumi about`

N/A

### Additional context

_No response_

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

Zaid-Ajaj:
Fixes #139 

When using an `async` lambda inside `Apply` where we `await` asynchronous functions that don't return anything, the lambda gets the signature `Func<T, Task>`. Giving that to `Apply` the returned `Task` is not awaited because the overload chosen for the `Apply` is the one accepting `Func<T, U>` and this overload doesn't know how to await and register `U` where `typeof(U) == typeof(Task)`

Because we are not registering the task, it is not awaited properly and the Pulumi program runs to completion before the task is resolved. 

We can easily repro this behavior as follows:
```cs
using System.Threading.Tasks;
using System;
using Pulumi;

await Deployment.RunAsync(() => 
{
    Output.Create(0).Apply(async _ => 
    {
        await LongFailingTask();
    });
});

async Task LongFailingTask()
{
    Pulumi.Log.Info("LongFailingTask before waiting");
    await Task.Delay(5000);
    Pulumi.Log.Info("LongFailingTask after waiting");
    throw new Exception("Boom!");
}
```
Running `pulumi up` on the code above will run the program but will never reach the exception after the delay:
```
Diagnostics:
  pulumi:pulumi:Stack (project-dev):
    LongFailingTask before waiting
```

The fix is to handle `typeof(U) == typeof(Task)` in a special way inside `Apply` where we create a wrapper task-returning function which _awaits_ the task provided by the user and returns a dummy empty task so that the signature `Func<T, Task>` still holds

### DIFF
--- a/sdk/Pulumi/Core/Output.cs
+++ b/sdk/Pulumi/Core/Output.cs
@@ -502,15 +502,15 @@ namespace Pulumi
         /// <summary>
         /// <see cref="Output{T}.Apply{U}(Func{T, Output{U}})"/> for more details.
         /// </summary>
-        public Output<Task> Apply(Func<T, Task> func)
+        public Output<ValueTuple> Apply(Func<T, Task> func)
         {
             return Apply(t =>
             {
-                async Task<Task> WrapperTask()
+                async Task<ValueTuple> WrapperTask()
                 {
                     var underlyingTask = func(t);
                     await underlyingTask.ConfigureAwait(false);
-                    return underlyingTask;
+                    return ValueTuple.Create();
                 }
 
                 return Output.Create(WrapperTask());
--- a/sdk/Pulumi/Core/Output.cs
+++ b/sdk/Pulumi/Core/Output.cs
@@ -503,7 +503,40 @@ namespace Pulumi
         /// <see cref="Output{T}.Apply{U}(Func{T, Output{U}})"/> for more details.
         /// </summary>
         public Output<U> Apply<U>(Func<T, U> func)
-            => Apply(t => Output.Create(func(t)));
+        {
+            if (typeof(U).GUID == typeof(Task).GUID)
+            {
+                // This is a special case for when the input lambda returns a Task implicitly because
+                // it is async lambda that doesn't return a value. This overload is resolved
+                // instead of the one accepting Task<T> and so the task doesn't get registered with the engine.
+                return Apply(t =>
+                {
+                    async Task<U> WrapperTask()
+                    {
+                        if (func(t) is Task underlyingTask)
+                        {
+                            await underlyingTask.ConfigureAwait(false);
+                        }
+
+                        // return an empty task of type U to satisfy the runtime
+                        // using reflection to create new Task(() => { })
+                        // which doesn't have to be awaited because we awaited the original task
+                        Action emptyAction = () => { };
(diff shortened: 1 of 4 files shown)


sdk/Pulumi/Core/Output.cs
```diff
@@ -499,11 +499,33 @@ internal static Output<T> CreateUnknown(T value)
         internal static Output<T> CreateUnknown(Func<Task<T>> valueFactory)
             => Unknown(default!).Apply(_ => valueFactory());
 
+        /// <summary>
+        /// <see cref="Output{T}.Apply{U}(Func{T, Output{U}})"/> for more details.
+        /// </summary>
+        public Output<Task> Apply(Func<T, Task> func)
```
Frassle, line 505: Oh this is slightly confusing isn't it. How I wish we had unit. The Task returned by the Output isn't the Task returned by the func, but a different CompletedTask. Which doesn't quite align with what `Apply(Func<Task<T>>)` does (the output returns the inner T).

I have seen the suggestion of using `System.ValueTuple` as a unit type for C#, maybe we should do that here?

Zaid-Ajaj: > The Task returned by the Output isn't the Task returned by the func, but a different CompletedTask.

Yeah not sure why I didn't return the awaited underlying task right away 😅 I guess a bit tired. Changed that ✅ 

> I have seen the suggestion of using System.ValueTuple as a unit type for C#, maybe we should do that here?

You mean the signature would become this?
```cs
Output<Task<ValueTuple> Apply(Func<T, Task> func)
```

Frassle: No I mean it would be `Output<ValueTuple> Apply(Func<T, Task> func)`, a poor mans `unit` basically. Imagine if we're in F# land we'd just have `Output<U> Apply<U>(Func<T, Task<U>> func)` and filled in for unit you'd get `Output<unit> Apply(Func<T, unit> func)`.

This would line up with how the `Task<T>` apply functions don't return the task, but the value inside the task. Just awkward with this one because there is no value, but unit doesn't exist.

Zaid-Ajaj: I actually like that a lot! Having `Output<ValueTuple>` also means we could even export this thing as an output, it seems that our serialization logic doesn't handle this though, would be nice to make it output `()`

Frassle (approved):

PR #192, opened 2023-10-22
Change changelog system to changie

Frassle:
We'll need to update https://github.com/pulumi/home/blob/master/infrastructure/github-automation/changelog-bot/index.ts for the changelog bot to understand this, but using a standard supported tool rather than our homebuilt go-change one seems sensible.


.github/workflows/release-dotnet-provider.yml
```diff
@@ -1,11 +1,7 @@
 on:
   push:
-    tags:
-      - v*.*.*
-    paths-ignore:
-      - "CHANGELOG.md"
-      - "CHANGELOG_PENDING.md"
-      - "README.md"
+    branches: [ main ]
+    paths: [ CHANGELOG.md ]
```
tgummerer, line 4: I think this makes sense, whenever changes to CHANGELOG.md get merged to main we'll create a new release.  

Would it also make sense to have the GitHub action tag the release automatically?  With the old process we had a guarantee that the release is tagged properly, as we'd only kick off the action when a new tag gets pushed, but there's no such guarantee anymore.

.goreleaser.yml
```diff
@@ -24,4 +24,7 @@ snapshot:
   name_template: "{{ .Tag }}-SNAPSHOT"
 
 checksum:
-  name_template: 'checksums.txt'
\ No newline at end of file
+  name_template: 'checksums.txt'
+
+changelog:
```
tgummerer, line 29: How does goreleaser know the tag?  Does it use the latest tag always?  If so we'd have to make sure to tag before merging CHANGELOG.md changes to main, or do some automated tagging as suggested above.

tgummerer (commented):
This is great!  I'd love to have a unified system for doing changelogs, learning about that in different repos was always a little bit more overhead than I liked.

.github/workflows/release-dotnet-provider.yml
Frassle: When goreleaser publishes the release to GitHub it makes a matching tag.

.goreleaser.yml
Frassle: We set the envvar `GORELEASER_CURRENT_TAG` to what changie tells us is the latest version (i.e. the one we just changed CHANGELOG.md for)

tgummerer (approved):

PR #200, opened 2023-10-31
[sdk] Implement reflection-based RegisterOutputs for component resources

Zaid-Ajaj:
When writing component resources, at the very end of the constructor logic, developers have to make a call to `RegisterOutputs(...)` and provide a bag of outputs from the instance properties declared with `[Output]` property. 

This PR makes it such that developers can use `RegisterOutputs()` without providing anything and it will use reflection to deduce the keys and values of the outputs from the properties of the component resource. It works as follows:
 - ~Any property of type `Output<T>` is considered an output with the key equal to the name of the property~
 - Any property marked with attribute `[Output]` is considered an output the key equal to the name of the property
 - Any property marked with attribute `[Output(name: "<overriddenPropertyName>")]` is considered an output the key equal to `<overriddenPropertyName>`

~Since we don't have a good way to test `RegisterOutputs`, I moved the logic to an internal, static function called `CollectOutputProperties` which is unit testable on its own. Then in a test stack, made it return the stack outputs from the collected output properties of a test component (`BasicComponent`) so we can assert the resulting keys and values~

The PR extends `IMocks` with support for `RegisterResourceOutputs` such that we can unit test the registered outputs by component resources or stacks

### DIFF
--- a/sdk/Pulumi.Tests/Resources/ComponentResourceTests.cs
+++ b/sdk/Pulumi.Tests/Resources/ComponentResourceTests.cs
@@ -103,6 +103,8 @@ namespace Pulumi.Tests.Resources
 
         class BasicComponent : ComponentResource
         {
+            [Output]
+            private Output<string> PrivateOutputProperty { get; set; }
             public Output<string> First { get; set; }
             [Output]
             public Output<string> Second { get; set; }
@@ -113,6 +115,9 @@ namespace Pulumi.Tests.Resources
                 First = Output.Create("first");
                 Second = Output.Create("second");
                 Third = Output.Create("third");
+                PrivateOutputProperty = Output.Create("private");
+                // RegisterOutputs will only register "Second" and "myThird" since
+                // those are public properties marked with [Output]
                 RegisterOutputs();
             }
         }
@@ -127,38 +132,29 @@ namespace Pulumi.Tests.Resources
                 new BasicComponent("basic");
             });
 
-            foreach (var resource in resources)
-            {
-                if (resource is BasicComponent basic)
-                {
-                    var urn = await basic.Urn.GetValueAsync("<unknown>");
-                    var outputs = mocks.GetRegisteredOutputs(urn) as IDictionary<string, object>;
-                    Assert.Contains("Second", outputs);
-                    Assert.Contains("myThird", outputs);
+            var basic = (BasicComponent)resources.First(resource => resource is BasicComponent);
 
-                    if (outputs["Second"] is Output<string> second)
-                    {
-                        var value = await second.GetValueAsync("<unknwon>");
-                        Assert.Equal("second", value);
-                    }
+            var urn = await basic.Urn.GetValueAsync("<unknown>");
+            var outputs = mocks.GetRegisteredOutputs(urn);
+            Assert.Equal(2, outputs.Count);
+            Assert.True(outputs.ContainsKey("Second"));
+            Assert.True(outputs.ContainsKey("myThird"));
 
-                    if (outputs["myThird"] is Output<string> third)
-                    {
-                        var value = await third.GetValueAsync("<unknwon>");
-                        Assert.Equal("third", value);
-                    }
-                }
-            }
+            var second = await outputs["Second"].GetValueAsync("<unknown>");
+            Assert.Equal("second", second);
+
+            var myThird = await outputs["myThird"].GetValueAsync("<unknown>");
+            Assert.Equal("third", myThird);
         }
     }
 
     class MinimalMocks : IMocks
     {
-        private readonly Dictionary<string, ImmutableDictionary<string, object>> _registeredOutputsByUrn;
+        private readonly Dictionary<string, ImmutableDictionary<string, Output<object?>>> _registeredOutputsByUrn;
 
         public MinimalMocks()
         {
-            _registeredOutputsByUrn = new Dictionary<string, ImmutableDictionary<string, object>>();
+            _registeredOutputsByUrn = new Dictionary<string, ImmutableDictionary<string, Output<object?>>>();
         }
 
         public Task<object> CallAsync(MockCallArgs args)
@@ -178,7 +174,7 @@ namespace Pulumi.Tests.Resources
             return Task.CompletedTask;
         }
 
-        public ImmutableDictionary<string, object> GetRegisteredOutputs(string urn)
+        public ImmutableDictionary<string, Output<object?>> GetRegisteredOutputs(string urn)
         {
             return _registeredOutputsByUrn[urn];
         }
--- a/sdk/Pulumi.Tests/Resources/ComponentResourceTests.cs
+++ b/sdk/Pulumi.Tests/Resources/ComponentResourceTests.cs
@@ -2,6 +2,8 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
 using System.Threading.Tasks;
 using Pulumi.Testing;
 using Xunit;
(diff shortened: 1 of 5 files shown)


Frassle:
> Since we don't have a good way to test RegisterOutputs

Really? Can't you test this in a mock stack?

sdk/Pulumi/Resources/ComponentResource.cs
```diff
@@ -52,6 +53,49 @@ public ComponentResource(string type, string name, ComponentResourceOptions? opt
...
+                {
+                    // otherwise of we only have [Output] we will simply use the name of the property itself
+                    // when exporting the value
+                    var value = prop.GetValue(instance);
+                    outputs.Add(prop.Name, value);
+                }
+                else
+                {
+                    var propertyType = prop.PropertyType;
+                    // only if the type is Output<T> then we will consider it an output property
```
Frassle, line 87: I'm inclined to not include these. Be specific that it's only things tagged [OutputAttribute] that get auto exported, a component is free to have non-Output\<T> things that are still "outputs", I don't think we should privilege the Output\<T> type.

sdk/Pulumi/Resources/ComponentResource.cs
```diff
@@ -52,6 +53,49 @@ public ComponentResource(string type, string name, ComponentResourceOptions? opt
             this.remote = remote;
         }
 
+        internal static Dictionary<string, object?> CollectOutputProperties(ComponentResource instance)
```
Frassle, line 56: I'd rather this stayed protected, or even better just folded this code into the RegisterOutputs() method. We should work out how to test this via a mock stack looking at resource outputs, that's what users will want to do.

sdk/Pulumi/Resources/ComponentResource.cs
Zaid-Ajaj: The reasoning here if a user has specified a property as `Output<T>` then it feels redundant to also mark it as `[Output]`. 

> a component is free to have non-Output<T> things that are still "outputs"

Is there any use for this? why have a field that is only local to the component 🤔

sdk/Pulumi/Resources/ComponentResource.cs
Zaid-Ajaj: > Really? Can't you test this in a mock stack?

Stacks already do their own reflection logic that calls `RegisterOutputs(outputs)` internally which is why I went with this.

sdk/Pulumi/Resources/ComponentResource.cs
Frassle: You might have a component that calculates a non-output result based on it's inputs. That result might be passed down to child resources but also returned to the user to be used later as well. So it's not "only local to the component" its still a public property, its just not Output-y.

sdk/Pulumi/Resources/ComponentResource.cs
Frassle: Sure, but users will want to test their component outputs via the mock test system so lets make sure that works.

sdk/Pulumi/Resources/ComponentResource.cs
Zaid-Ajaj: Alright that makes sense ✅ I've updated the function and now only properties marked with `[Output]` are considered outputs of the component to be registered

sdk/Pulumi/Resources/ComponentResource.cs
Zaid-Ajaj: I've extended `IMocks` now to support `RegisterResourceOutputs` which makes the code unit testable the way users would want to test their outputs. Folder the code back into `RegisterOutputs()` and refactored the unit test.

> `IMocks` has a default implementation of `RegisterResourceOutputs` that returns `Task.CompletedTask` so it doesn't break existing implementations

sdk/Pulumi/Testing/IMocks.cs
```diff
@@ -78,4 +84,19 @@ public class MockCallArgs
...
+
+    /// <summary>
+    /// MockRegisterResourceOutputsRequest for use in RegisterOutputRequest
+    /// </summary>
+    public class MockRegisterResourceOutputsRequest
+    {
+        /// <summary>
+        /// The URN of the resource of which the outputs are being registered
+        /// </summary>
+        public string? Urn { get; set; }
```
Frassle, line 96: Is this really nullable? Don't we all ways have an URN?

sdk/Pulumi/Testing/IMocks.cs
```diff
@@ -78,4 +84,19 @@ public class MockCallArgs
...
+        /// <summary>
+        /// The outputs which have been registered by the resource
+        /// </summary>
+        public ImmutableDictionary<string, object> Outputs { get; set; } = null!;
```
Frassle, line 100: Why ` = null!`?

sdk/Pulumi/Resources/ComponentResource.cs
```diff
@@ -61,7 +62,38 @@ public ComponentResource(string type, string name, ComponentResourceOptions? opt
         /// state as quickly as possible (instead of waiting until the entire application completes).
         /// </summary>
         protected void RegisterOutputs()
-            => RegisterOutputs(ImmutableDictionary<string, object?>.Empty);
+        {
+            var outputs = new Dictionary<string, object?>();
+            var currentType = GetType();
+            foreach (var prop in currentType.GetProperties())
```
Frassle, line 68: GetProperties only returns public props right?

sdk/Pulumi/Resources/ComponentResource.cs
```diff
@@ -61,7 +62,38 @@ public ComponentResource(string type, string name, ComponentResourceOptions? opt
...
+                if (prop.Name == nameof(Urn))
+                {
+                    continue;
+                }
+
+                var outputAttribute =
+                    prop.GetCustomAttributes(typeof(OutputAttribute), false)
+                        .FirstOrDefault();
+
+                if (outputAttribute is OutputAttribute attr && attr.Name != null)
```
Frassle, line 79: These two blocks are the same _except_ for the name used. Maybe refactor into something like:
```
if is OutputAttribute {
  name = prop.Name
  if attr.Name != null {
    name = attr.Name
  }
  var value = prop.GetValue(this)
  outputs.Add(name, value)
}
```

sdk/Pulumi.Tests/Resources/ComponentResourceTests.cs
```diff
@@ -112,5 +171,16 @@ public Task<object> CallAsync(MockCallArgs args)
...
             return (args.Name + "-id", args.Inputs);
         }
+
+        public Task RegisterResourceOutputs(MockRegisterResourceOutputsRequest request)
+        {
+            _registeredOutputsByUrn[request.Urn ?? ""] = request.Outputs;
+            return Task.CompletedTask;
+        }
+
+        public ImmutableDictionary<string, object> GetRegisteredOutputs(string urn)
```
Frassle, line 181: Should we just give a read-only view access to the whole dictionary? Then users can either decided to lookup by URN or just iterate over all the resources.

sdk/Pulumi.Tests/Resources/ComponentResourceTests.cs
```diff
@@ -98,10 +100,67 @@ public async void PropagatesProvidersOption()
...
             });
         }
+
+        class BasicComponent : ComponentResource
+        {
+            public Output<string> First { get; set; }
+            [Output]
+            public Output<string> Second { get; set; }
+            [Output("myThird")]
+            public Output<string> Third { get; set; }
```
Frassle, line 110: Maybe add a private [Output] to show that _doesn't_ export?

sdk/Pulumi.Tests/Resources/ComponentResourceTests.cs
```diff
@@ -98,10 +100,67 @@ public async void PropagatesProvidersOption()
...
+            foreach (var resource in resources)
+            {
+                if (resource is BasicComponent basic)
+                {
+                    var urn = await basic.Urn.GetValueAsync("<unknown>");
+                    var outputs = mocks.GetRegisteredOutputs(urn) as IDictionary<string, object>;
+                    Assert.Contains("Second", outputs);
+                    Assert.Contains("myThird", outputs);
+
+                    if (outputs["Second"] is Output<string> second)
```
Frassle, line 139: Can't we just cast here, if the cast fails it will just fail the test but we expect that. Also catches if we accidentally change this so outputs["second"] becomes null or string.

sdk/Pulumi/Testing/IMocks.cs
Zaid-Ajaj: Was following the example above it but I changed the class into a record

sdk/Pulumi/Resources/ComponentResource.cs
Zaid-Ajaj: Yes, the unit test now shows that a private property isn't registered

sdk/Pulumi/Resources/ComponentResource.cs
Zaid-Ajaj: That's nicer! updated the implementation ✅

sdk/Pulumi.Tests/Resources/ComponentResourceTests.cs
Zaid-Ajaj: This is specific to the unit test in question and how it implemented `IMocks`, nothing user-facing

sdk/Pulumi.Tests/Resources/ComponentResourceTests.cs
Zaid-Ajaj: Changed the dictionary to `string -> Output<object?>` in the implementation because the deserializer from `Struct` always gives `OutputData<object?>` which makes casting only needed at the value level after awaiting it.

sdk/Pulumi.Tests/Resources/ComponentResourceTests.cs
Frassle: Ah yes, my oversight, that's all good then.

Frassle (approved):

PR #205, opened 2023-11-15
Run tests on dotnet 8

Frassle:
Keep running test on 6 as we still support that. Honestly these tests _should_ have been running on 6 and 7 already.


tgummerer (approved):
Should we include 7.0.x in the matrix as well just in case?  Or do we expect that to work either way if 6.0.x and 8.0.x work?

Frassle:
> Should we include 7.0.x in the matrix as well just in case? Or do we expect that to work either way if 6.0.x and 8.0.x work?

Yes, it can get very expensive to test every possible version so we generally just go with oldest and newest.

PR #210, opened 2023-12-14
Add environment add and remove commands to automation api

cleverguy25:
Adds support for environment add/rm commands to dotnet.

### DIFF
--- a/sdk/Pulumi.Automation/Workspace.cs
+++ b/sdk/Pulumi.Automation/Workspace.cs
@@ -124,6 +124,25 @@ namespace Pulumi.Automation
         public abstract Task PostCommandCallbackAsync(string stackName, CancellationToken cancellationToken = default);
 
         /// <summary>
+        /// Adds environments to the end of a stack's import list. Imported environments are merged in order
+        /// per the ESC merge rules. The list of environments behaves as if it were the import list in an anonymous
+        /// environment.
+        /// </summary>
+        /// <param name="stackName">The name of the stack.</param>
+        /// <param name="environments">List of environments to add to the end of the stack's import list.</param>
+        /// <param name="cancellationToken">A cancellation token.</param>
+        public abstract Task AddEnvironmentsAsync(string stackName, IEnumerable<string> environments, CancellationToken cancellationToken = default);
+
+
+        /// <summary>
+        /// Removes environments from a stack's import list.
+        /// </summary>
+        /// <param name="stackName">The name of the stack.</param>
+        /// <param name="environment">The name of the environment to remove from the stack's configuration.</param>
+        /// <param name="cancellationToken">A cancellation token.</param>
+        public abstract Task RemoveEnvironmentAsync(string stackName, string environment, CancellationToken cancellationToken = default);
+
+        /// <summary>
         /// Returns the value associated with the specified stack name and key,
         /// scoped to the Workspace.
         /// </summary>
--- a/.changes/unreleased/Improvements-210.yaml
+++ b/.changes/unreleased/Improvements-210.yaml
@@ -1,0 +1,6 @@
+component: sdk
+kind: Improvements
+body: Add environment add and remove commands to automation api
+time: 2023-12-13T17:49:09.172346903-08:00
+custom:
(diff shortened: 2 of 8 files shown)


komalali (approved):
LGTM but would be great to get these tests running in CI and an eye from the repo maintainers.

Frassle (commented):
SDK changes look fine, will leave for Justin to comment on getting the tests running

sdk/Pulumi.Automation/Workspace.cs
```diff
@@ -123,6 +123,25 @@ internal Workspace(IPulumiCmd cmd)
...
 
+        /// <summary>
+        /// Adds environments to the end of a stack's import list. Imported environments are merged in order
+        /// per the ESC merge rules. The list of environments behaves as if it were the import list in an anonymous
+        /// environment.
+        /// </summary>
+        /// <param name="stackName">The name of the stack.</param>
+        /// <param name="environments">List of environments to add to the end of the stack's import list.</param>
+        /// <param name="cancellationToken">A cancellation token.</param>
+        public abstract Task AddEnvironmentsAsync(string stackName, IEnumerable<string> environments, CancellationToken cancellationToken = default);
```
justinvp, line 134: The new APIs in the other SDKs use varargs for the `environments` arg ([Node](https://github.com/pulumi/pulumi/blob/6ae7c39498cc476d9240d93d52c9bd79cfcafd4b/sdk/nodejs/automation/workspace.ts#L102), [Python](https://github.com/pulumi/pulumi/blob/6ae7c39498cc476d9240d93d52c9bd79cfcafd4b/sdk/python/lib/pulumi/automation/_workspace.py#L210), and [Go](https://github.com/pulumi/pulumi/blob/6ae7c39498cc476d9240d93d52c9bd79cfcafd4b/sdk/go/auto/workspace.go#L48)):

```typescript
// Usage: await addEnvironments("stackname", "environment1", "environment2");
addEnvironments(stackName: string, ...environments: string[]): Promise<void>;
```

```python
# Usage: add_environment("stackname", "environment1", "environment2")
def add_environments(self, stack_name: str, *environment_names: str) -> None:
```

```go
// Usage: AddEnvironment(ctx, "stackname", "environment1", "environment2")
AddEnvironments(context.Context, string, ...string) error
```

We could add an overload that accepted a `params` array to make it similarly more ergonomic to use in most cases when you don't care about passing a cancellation token:

```c#
        /// <summary>
        /// Adds environments to the end of a stack's import list. Imported environments are merged in order
        /// per the ESC merge rules. The list of environments behaves as if it were the import list in an anonymous
        /// environment.
        /// </summary>
        /// <param name="stackName">The name of the stack.</param>
        /// <param name="environments">List of environments to add to the end of the stack's import list.</param>
        public Task AddEnvironmentsAsync(string stackName, params string[] environments)
            => AddEnvironmentsAsync(stackName, (IEnumerable<string>)environments);
```

Then usage in most cases would look like:

```c#
await workspace.AddEnvironmentsAsync(stackName, "automation-api-test-env", "automation-api-test-env-2");
```

Rather than:

```c#
await workspace.AddEnvironmentsAsync(stackName, new[] { "automation-api-test-env", "automation-api-test-env-2" });
```

And if we add it here, add one to `WorkspaceStack` as well.

Aside: We could likely do that for other cases where we accept an `IEnumerable<string>` arg.

justinvp (approved):
Some suggestions, but LGTM. I'd be fine with merging this, and we can sort out the CI issue separately.

@Frassle, do we need to keep `sdk/Pulumi.Automation/PublicAPI.Shipped.txt` up-to-date? If so, is there an easy way to update it. I always find it to be painful because I've done it by hand in the past.

sdk/Pulumi.Automation/Workspace.cs
cleverguy25: @justinvp I agree with all of this and will follow up. Any reservation on that coming in a follow up PR and merging this PR?

justinvp: Follow-up PR works for me!

PR #217, opened 2023-12-27
Upgrade to pulumi v3.99.0

justinvp:
Also update Go to v1.21.


Frassle:
You need to fill in the codegen grpc methods with stub NotImplemented.

Issue #218, opened 2024-01-04
Update `Grpc.AspNetCore.Server.Reflection` to a more recent version


pierskarsenbarg:
Customer has pointed out that we're on [v2.37.0 of this nuget package](https://github.com/pulumi/pulumi-dotnet/blob/main/sdk/Pulumi/Pulumi.csproj#L35) and it's throwing alerts in their static analysis tooling.

The version we're using is dated 20/04/2021 so it is nearly [3 years old now ](https://www.nuget.org/packages/Grpc.AspNetCore.Server.Reflection/2.37.0#versions-body-tab)

![DependencyTree](https://github.com/pulumi/pulumi-dotnet/assets/109047/c5e8e384-5f69-49d4-987e-cb71f9a59e9b)
![Vulnerability](https://github.com/pulumi/pulumi-dotnet/assets/109047/18efb036-e28d-44fe-af5e-44e2d221a009)

ttvrdon:
@pierskarsenbarg Thanks Pier for really quick action with this fix. I reviewed the PR and this will not help...
Vulnerability was fixed in version 2.59.0

Frassle:
That's going to be trickier because it drops support for dotnet 3. I'll talk to the team about what we can do here.

mikhailshilkov:
@Frassle Why is dropping dotnet 3 a problem? It's not a supported version for 1.5 years, right?

Frassle:
> Why is dropping dotnet 3 a problem?

We broke a load of downstream provider builds last time, it was technically a few months out of support even then. It _should_ be fine, but I haven't been able to get anyone to confirm it's ok to make this break change.

PR #219, opened 2024-01-04
Update Grpc dependencies

Frassle:
Fixes https://github.com/pulumi/pulumi-dotnet/issues/218.

2.51.0 seems to be the latest release we can safely update to.
2.52 and later of Grpc.Tools doesn't work on old macos versions (https://github.com/grpc/grpc/issues/32558).
2.57 and later Grpc libraries dropped support for netcore3 which we're still building for, although _technically_ it's well out of support so we ought to drop it at some point but we hit lots of pain last time we tried (https://github.com/pulumi/pulumi/issues/11768).

### DIFF
--- a/sdk/Pulumi/Pulumi.csproj
+++ b/sdk/Pulumi/Pulumi.csproj
@@ -30,10 +30,10 @@
 
   <ItemGroup>
     <PackageReference Include="Ben.Demystifier" Version="0.4.1" />
-    <PackageReference Include="Grpc.Net.Client" Version="2.52.0" />
-    <PackageReference Include="Grpc.AspNetCore.Server" Version="2.37.0" />
-    <PackageReference Include="Grpc.AspNetCore.Server.Reflection" Version="2.37.0" />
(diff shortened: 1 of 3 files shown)


tgummerer (approved):

sdk/Pulumi/Pulumi.csproj
```diff
@@ -30,10 +30,10 @@
 
   <ItemGroup>
     <PackageReference Include="Ben.Demystifier" Version="0.4.1" />
-    <PackageReference Include="Grpc.Net.Client" Version="2.52.0" />
-    <PackageReference Include="Grpc.AspNetCore.Server" Version="2.37.0" />
-    <PackageReference Include="Grpc.AspNetCore.Server.Reflection" Version="2.37.0" />
-    <PackageReference Include="Grpc.Tools" Version="2.44.0">
+    <PackageReference Include="Grpc.Net.Client" Version="2.51.0" />
```
ttvrdon, line 33: This will not help...
Reported security vulnerability was fixed in v 2.59.0

PR #227, opened 2024-01-29
Run build & test steps on windows & macos if the ci/test label is set

julienp:
Currently we only run a [certain integration tests](https://github.com/pulumi/pulumi-dotnet/blob/a5f3d6cefbab84efacd4df74a21aac8d3f3b3d2f/.github/workflows/pr.yml#L67) on windows & Mac.

Allow running the sdk and automation tests on windows & Mac by setting the `ci/test` label.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -24,6 +24,16 @@ jobs:
           os="${{ contains(github.event.pull_request.labels.*.name, 'ci/test') && 'ubuntu-latest macos-latest windows-latest' || 'ubuntu-latest' }}"
           echo "matrix={\"os\": $(echo $os | jq -cR 'split(" ")')}" >> $GITHUB_OUTPUT
 
+  format:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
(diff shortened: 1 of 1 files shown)


.github/workflows/pr.yml
```diff
@@ -14,9 +14,33 @@ env:
...
+        run: echo XX"${{ fromJson(needs.setup_matrix.outputs.platforms) }}"XX
+
   build:
+    needs: setup_matrix
     strategy:
       matrix:
+        os: ${{ fromJson(needs.setup_matrix.outputs.platforms) }}
         dotnet-version: [6.0.x, 8.0.x]
 
     runs-on: ubuntu-latest
```
Frassle, line 46: Doesn't this need to change to ${{ matrix.os }}

tgummerer (approved):
I think https://github.com/pulumi/pulumi-dotnet/pull/227/files#diff-15c806aa509538190832852f439e9921a23bec2da81f95ed0e4bf13c14e5b160R146 now also needs a dependency on `format`.  Other than that this LGTM!

PR #228, opened 2024-01-31
Publish dev SDK

tgummerer:
Publish a dev SDK for pulumi-dotnet which we can consume later for pre-release testing.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -122,6 +122,40 @@ jobs:
         run: dotnet run integration test TestProvider
       - name: TestDeletedWith
         run: dotnet run integration test TestDeletedWith
+
+  info:
+    name: gather
+    runs-on: ubuntu-latest
+    outputs:
+      version: "${{ steps.version.outputs.version }}"
+    steps:
+      - name: Checkout code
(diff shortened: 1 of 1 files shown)


.github/workflows/pr.yml
```diff
@@ -122,6 +122,35 @@ jobs:
...
+      version: "${{ steps.version.outputs.version }}"
+    steps:
+      - uses: actions/checkout@v4
+      - name: Git describe
+        id: ghd
+        uses: proudust/gh-describe@v1
+      - name: Stitch version together
+        id: version
+        run: |
+          TAG="${{ steps.ghd.outputs.tag }}"
```
Frassle, line 139: Do we not want to make use of the current version info like the normal release does? 
https://github.com/pulumi/pulumi-dotnet/blob/main/.github/workflows/release-dotnet-provider.yml#L34C1-L38C23

I'd expect this to query changie for the _next_ version and then add a dev hash to the end of that.

tgummerer: Yeah good point, I've changed it to using changie.

Frassle (approved):
Looks about right

PR #231, opened 2024-02-17
Add attribute to handle deserialization of constructor parameters with name overrides

praneetloke:
This PR is in support of https://github.com/pulumi/pulumi/issues/14130.

I have a local override for the native provider in question to use the changes in this PR, as well as codegen changes in pulumi/pulumi using the PR https://github.com/pulumi/pulumi/pull/14246 with additional changes to emit the attribute introduced in this PR.

Here's how it looks in practice in a provider's SDK:

```csharp
    [OutputType]
    public sealed class ListOneClicksProperties
    {
        public readonly ImmutableArray<Outputs.OneClicks> _1Clicks;

        [OutputConstructor]
        private ListOneClicksProperties([OutputConstructorParameterAttribute("1_clicks")] ImmutableArray<Outputs.OneClicks> _1Clicks)
        {
            this._1Clicks = _1Clicks;
        }
    }
```

### DIFF
--- a/sdk/Pulumi/Serialization/Attributes.cs
+++ b/sdk/Pulumi/Serialization/Attributes.cs
@@ -114,13 +114,13 @@ namespace Pulumi
     /// constructor parameters with a name override.
     /// </summary>
     [AttributeUsage(AttributeTargets.Parameter)]
-    public class ConstructorParameterNameOverrideAttribute : Attribute
+    public sealed class OutputConstructorParameterAttribute : Attribute
     {
         public string Name { get; }
 
-        public ConstructorParameterNameOverrideAttribute(string apiName)
+        public OutputConstructorParameterAttribute(string name)
         {
-            Name = apiName;
+            Name = name;
         }
     }
 }
--- a/sdk/Pulumi/Serialization/Attributes.cs
+++ b/sdk/Pulumi/Serialization/Attributes.cs
@@ -108,4 +108,19 @@ namespace Pulumi
             Version = version;
         }
     }
+
+    /// <summary>
+    /// Attribute used by a Pulumi Cloud Provider Package to mark constructor parameters
+    /// whose API name is different from the C# name.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Parameter)]
+    public class ConstructorParameterApiNameAttribute : Attribute
+    {
+        public string ApiName { get; }
+
+        public ConstructorParameterApiNameAttribute(string apiName)
+        {
+            ApiName = apiName;
+        }
+    }
 }
--- a/sdk/Pulumi/Serialization/Attributes.cs
+++ b/sdk/Pulumi/Serialization/Attributes.cs
@@ -78,6 +78,21 @@ namespace Pulumi
     }
 
     /// <summary>
+    /// Attribute used by a Pulumi Cloud Provider Package to mark
+    /// constructor parameters with a name override.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Parameter)]
+    public sealed class OutputConstructorParameterAttribute : Attribute
+    {
+        public string Name { get; }
+
+        public OutputConstructorParameterAttribute(string name)
+        {
+            Name = name;
+        }
+    }
+
+    /// <summary>
     /// Attribute used by a Pulumi Cloud Provider Package to mark enum types.
     ///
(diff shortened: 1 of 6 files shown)


sdk/Pulumi/Serialization/Attributes.cs
```diff
@@ -108,4 +108,19 @@ public ResourceTypeAttribute(string type, string? version)
             Version = version;
         }
     }
+
+    /// <summary>
+    /// Attribute used by a Pulumi Cloud Provider Package to mark
+    /// constructor parameters with a name override.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Parameter)]
+    public class ConstructorParameterNameOverrideAttribute : Attribute
```
justinvp, line 117: The type should be `sealed`.

Also, I wonder if a better name would be `OutputConstructorParameterAttribute`, to align with `OutputConstructorAttribute`:

```suggestion
    public sealed class OutputConstructorParameterAttribute : Attribute
```

I also wonder if this should be moved directly below `OutputConstructorAttribute` (line 76).

sdk/Pulumi/Serialization/Attributes.cs
```diff
@@ -108,4 +108,19 @@ public ResourceTypeAttribute(string type, string? version)
...
+    {
+        public string Name { get; }
+
+        public ConstructorParameterNameOverrideAttribute(string apiName)
+        {
+            Name = apiName;
+        }
```
justinvp, line 124:
```suggestion
        public ConstructorParameterNameOverrideAttribute(string name)
        {
            Name = name;
        }
```

sdk/Pulumi/Serialization/Converter.cs
```diff
@@ -200,11 +200,18 @@ private static (object?, string?) TryConvertObject(Action<string> warn, string c
             for (int i = 0, n = constructorParameters.Length; i < n; i++)
             {
                 var parameter = constructorParameters[i];
+                var parameterName = parameter.Name!;
+                var attribute = parameter.GetCustomAttribute<ConstructorParameterNameOverrideAttribute>();
+
+                if (attribute != null)
```
justinvp, line 206: Nit

```suggestion
                if (attribute != null)
```

justinvp (commented):
Thanks @praneetloke! Some comments.

Also, this needs tests. See https://github.com/pulumi/pulumi-dotnet/blob/8f1dd5e358a18aa8ac3f034dcb22818f2fa77e36/sdk/Pulumi.Tests/Serialization/ComplexTypeConverterTests.cs#L82

Your usage in the PR description doesn't quite look how I'd expect it to. I'd expect something like:

```c#
    [OutputType]
    public sealed class ListOneClicksProperties
    {
        public readonly ImmutableArray<Outputs.OneClicks> _1Clicks;

        [OutputConstructor]
        private ListOneClicksProperties([OutputConstructorParameter("1_clicks")] ImmutableArray<Outputs.OneClicks> _1Clicks)
        {
            this._1Clicks = _1Clicks;
        }
    }
```

sdk/Pulumi/Serialization/Attributes.cs
praneetloke: I've made changes for both suggestions you've made here. Also on the topic of the attribute class being `sealed` I noticed that `ResourceTypeAttribute` is not marked as such. That's the one I was using as reference and hence didn't mark the new one as `sealed` initially. Would you like me to mark it as `sealed` as well?

praneetloke:
@justinvp I've pushed new commits with changes based on your feedback.

> Your usage in the PR description doesn't quite look how I'd expect it to. I'd expect something like:

Ah, indeed. I made a mistake in the codegen and wrote out the attribute on the class member instead of the constructor param...despite the name of the attribute 🤦‍♂️ . I'll fix that when I open the PR in `pulumi/pulumi` though. I've noted it down.

praneetloke:
Bump.

justinvp (approved):
LGTM! Thank you @praneetloke and apologies for the long delay getting this reviewed!

praneetloke:
@justinvp awesome! The status checks for this PR still says that the workflow requires approval from a maintainer though and haven't run. Any idea why that might be?

Frassle:
Looks like you just need to run "dotnet run format-sdk"

praneetloke:
Ah got it. I'll remember that for next time. Thanks, @Frassle.

PR #233, opened 2024-03-01
Update proto files

Frassle:
Update proto files to latest published from pulumi to get access to new protocol features such as transforms.


proto/pulumi/analyzer.proto
justinvp, line 1: I didn't realize we copy these.

justinvp (approved):
LGTM

I get that the PR title is straightforward but the PR really should still have some kind of description with a bit more detail, even if it's only an additional sentence or two, explaining some additional context and why the change is needed.

proto/pulumi/analyzer.proto
Frassle: Yeh the choices for sharing proto files isn't great. This, or a git submodule or git subtree. But always ends up being some sort of copy.

PR #245, opened 2024-03-26
automation: always read complete lines

tgummerer:
This is the equivalent as https://github.com/pulumi/pulumi/pull/15778 for dotnet.  When tailing the event log we currently don't make sure that we read complete lines.  Lines can be flushed in the middle by the OS while they are being written down, leading to incomplete JSON.

The JSON encoder always ends each line with a newline, so we can stitch the lines back together even if we occasionally read a partly written line.


tgummerer:
Closing this for now, since I can't confirm this is an actual issue in dotnet.

PR #247, opened 2024-03-26
required output can be mocked

jasond-s:
fixes the same issue as seen in issue #9212 for the typescript sdk.

### DIFF
--- a/sdk/Pulumi/Resources/StackReference.cs
+++ b/sdk/Pulumi/Resources/StackReference.cs
@@ -71,12 +71,11 @@ namespace Pulumi
         public Output<object> RequireOutput(Input<string> name)
         {
             var inputs = (Input<ImmutableDictionary<string, object>>)this.Outputs;
-            var stackName = (Input<string>)this.Name;
-            var value = Output.Tuple(name, stackName, inputs).Apply(v =>
-                v.Item3.TryGetValue(v.Item1, out var result)
+            var value = Output.Tuple(name, inputs).Apply(v =>
+                v.Item2.TryGetValue(v.Item1, out var result)
                     ? result
                     : throw new KeyNotFoundException(
-                        $"Required output '{v.Item1}' does not exist on stack '{v.Item2}'."));
+                        $"Required output '{v.Item1}' does not exist on stack '{base.GetResourceName()}'."));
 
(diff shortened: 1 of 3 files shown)


sdk/Pulumi/Resources/StackReference.cs
```diff
@@ -71,12 +71,11 @@ public StackReference(string name, StackReferenceArgs? args = null, CustomResour
...
             var inputs = (Input<ImmutableDictionary<string, object>>)this.Outputs;
-            var stackName = (Input<string>)this.Name;
-            var value = Output.Tuple(name, stackName, inputs).Apply(v =>
-                v.Item3.TryGetValue(v.Item1, out var result)
+            var value = Output.Tuple(name, inputs).Apply(v =>
+                v.Item2.TryGetValue(v.Item1, out var result)
                     ? result
                     : throw new KeyNotFoundException(
-                        $"Required output '{v.Item1}' does not exist on stack '{v.Item2}'."));
+                        $"Required output '{v.Item1}' does not exist on stack '{base.GetResourceName()}'."));
```
Frassle, line 78: This will return the logical name of the stack reference resource rather than the name of the stack. Given the phrasing of the message that's probably not what's really wanted?
I think the issue here is that FakeStackOutputMocks doesn't return the name output.

Frassle: Yup raised https://github.com/pulumi/pulumi-dotnet/pull/260 to add a test showing how to do this. Just requires the mock to return the "name" output as well as the "outputs" for the stack.

jasond-s: Thanks, this test example is much clearer. I think I missed it as I was debugging through looking for `.Name` for resources and not `.Type` is there any documentation on the meaning of these different props on the args to mocks?

I am still getting a lot of errors in my test code, but I think they are probably just more missing mocks.

Frassle: > is there any documentation on the meaning of these different props on the args to mocks?

Not that I know of, certainly doesn't look like anything on https://www.pulumi.com/docs/using-pulumi/testing/unit/#add-mocks. Worth raising a bug to the docs repo (https://github.com/pulumi/docs) asking for more written on this.

PR #250, opened 2024-03-30
Add more FSharp Ops helpers

haodeon:
Since F# 6 improvements to implicit type conversions, it has been much more enjoyable to work with Pulumi.

Though I find there are still a few situations where it's not as easy as it could be, for me they were often properties with `InputList` type.

Adding these to the Ops module would make life easier.


Frassle (approved):

Frassle:
These look reasonable, thanks.

PR #251, opened 2024-04-11
Handle Outputs in derived Stacks

starkpl:
I want to be able to write a base class for a stack, which will be reused for a few projects which have separate, but similar infrastructures. Currently, the Output properties defined on the base class are not being registered in the derived stacks because of the `BindingFlags.DeclaredOnly` flag used in the `Stack.RegisterPropertyOutputs` method.

In this PR I remove this flag. Instead, I explicitly cut off properties from ancestors of the `Stack` class (currently the `Resource` class has `Urn` output property and I believe this was the reason to add the `DeclaredOnly` flag in the first place).

Check the test I added to see exactly what I wanted to achieve.

### DIFF
--- a/.changes/unreleased/Improvements-251.yaml
+++ b/.changes/unreleased/Improvements-251.yaml
@@ -1,0 +1,6 @@
+component: sdk
+kind: Improvements
+body: Handle Outputs in derived Stacks
+time: 2024-04-11T11:10:42.227518+02:00
+custom:
+  PR: "251"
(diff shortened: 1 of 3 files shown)


Frassle (approved):
Looks reasonable to me, nice fix.

starkpl:
@Frassle thanks for the approve. It looks like some checks failed in the Merge Queue (same that passed in the PR, doesn't look correlated with my changes). Any idea what happened and how can we merge this PR?

Frassle:
Odd, we'll try another merge if that fails I'll dig into what's up. The PR test passed so it probably isn't due to this change.

Frassle:
Yeh that test is failing on main now, something must have changed in the CLI. Raised https://github.com/pulumi/pulumi-dotnet/issues/252 to fix it.

starkpl:
Thanks, can't wait to use it 👍

PR #257, opened 2024-04-16
Changelog for v3.61.0

Frassle:


tgummerer (approved):

tgummerer:
This seems like another benefit of "changie" FWIW.  It's easy to go through and review all the changelog entries and comment on them again before the release gets frozen, and even potentially edit them.  At least the editing/suggesting changes part is a bit harder in the pulumi/pulumi freeze PRs.

PR #259, opened 2024-04-16
Make dotnet SDK support result reporting

tgummerer:
Similar to what we're doing to the other SDKs in https://github.com/pulumi/pulumi/pull/15740, this enables dealing with the Result field in the RegisterResource response.

### DIFF
--- a/sdk/Pulumi/Deployment/Deployment_ReadResource.cs
+++ b/sdk/Pulumi/Deployment/Deployment_ReadResource.cs
@@ -9,7 +9,7 @@ namespace Pulumi
 {
     public partial class Deployment
     {
-        private async Task<(string urn, string id, Struct data, ImmutableDictionary<string, ImmutableHashSet<Resource>> dependencies)> ReadResourceAsync(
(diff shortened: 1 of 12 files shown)


sdk/Pulumi/Deployment/Deployment_ReadResource.cs
```diff
@@ -42,7 +42,7 @@ public partial class Deployment
             // Now run the operation, serializing the invocation if necessary.
             var response = await this.Monitor.ReadResourceAsync(resource, request).ConfigureAwait(false);
 
-            return (response.Urn, id, response.Properties, ImmutableDictionary<string, ImmutableHashSet<Resource>>.Empty);
+            return (response.Urn, id, response.Properties, ImmutableDictionary<string, ImmutableHashSet<Resource>>.Empty, Pulumirpc.Result.Success);
```
tgummerer, line 45: `ReadResource` doesn't have a result field, so we always pretend it's a success.  Not sure if we should pull read and register resource apart here instead of unifying them in a single method.

Frassle (approved):
Looks fine to me, we'll get conformance tests in soon that will check this as well.

PR #263, opened 2024-04-22
Pin Google.Protobuf to 3.24

Frassle:
Fixes https://github.com/pulumi/pulumi-dotnet/issues/249.

We should get conformance tests up and running to confirm large request/responses to and from the engine work ok. For now we can just pin as the old version seemed to work.hw

### DIFF
--- a/sdk/Pulumi/Pulumi.csproj
+++ b/sdk/Pulumi/Pulumi.csproj
@@ -29,6 +29,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Google.Protobuf" Version="3.24.0" />
     <PackageReference Include="Ben.Demystifier" Version="0.4.1" />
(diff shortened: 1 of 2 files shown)


sdk/Pulumi/Pulumi.csproj
```diff
@@ -29,6 +29,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Google.Protobuf" Version="3.24.0" />
```
tgummerer, line 32: Just for my own understanding (because I don't know much about dotnet), was this dependency just missing before, or is it not expected for dotnet projects to specify all its dependencies, but only the pinned ones?

tgummerer (approved):

sdk/Pulumi/Pulumi.csproj
Frassle: No need to specify transitive dependencies unless you need to change the version range.

PR #266, opened 2024-04-24
Revert "Update Grpc dependencies (#219)"

Frassle:
This reverts commit d87f6a0f9d4780ed27bea8f97a7bff6a73a27914.

Fixes https://github.com/pulumi/pulumi-dotnet/issues/249.

### DIFF
--- a/sdk/Pulumi/Pulumi.csproj
+++ b/sdk/Pulumi/Pulumi.csproj
@@ -31,10 +31,10 @@
   <ItemGroup>
     <PackageReference Include="Google.Protobuf" Version="3.24.0" />
     <PackageReference Include="Ben.Demystifier" Version="0.4.1" />
-    <PackageReference Include="Grpc.Net.Client" Version="2.51.0" />
-    <PackageReference Include="Grpc.AspNetCore.Server" Version="2.51.0" />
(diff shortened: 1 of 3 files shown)


sdk/Pulumi/Pulumi.csproj
```diff
@@ -31,10 +31,10 @@
   <ItemGroup>
     <PackageReference Include="Google.Protobuf" Version="3.24.0" />
     <PackageReference Include="Ben.Demystifier" Version="0.4.1" />
-    <PackageReference Include="Grpc.Net.Client" Version="2.51.0" />
-    <PackageReference Include="Grpc.AspNetCore.Server" Version="2.51.0" />
-    <PackageReference Include="Grpc.AspNetCore.Server.Reflection" Version="2.51.0" />
-    <PackageReference Include="Grpc.Tools" Version="2.51.0">
+    <PackageReference Include="Grpc.Net.Client" Version="2.52.0" />
```
tgummerer, line 34: I was curious why this was upgraded while all the other libraries are getting downgraded.  But it looks like this got downgraded for some reason when upgrading gRPC, so it looks like the right thing to do.

tgummerer (approved):

PR #275, opened 2024-06-06
Resolves #46 Support authoring multi-language components in .NET

### LINKED ISSUE #46: Support authoring multi-language components in .NET
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

pulumi/pulumi#5485 adds support for using multi-language components from .NET. This issue tracks being able to author multi-language components in .NET.

algompluecker:


sdk/Pulumi/Provider/PropertyDependencies.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System.Collections.Generic;
+
+namespace Pulumi.Experimental.Provider;
+
+public sealed class PropertyDependencies
```
Frassle, line 5: Generally wouldn't bother adding a new type just to wrap a single field.

sdk/Pulumi/Provider/Provider.cs
```diff
@@ -916,14 +966,156 @@ public override async Task<Empty> Delete(Pulumirpc.DeleteRequest request, Server
...
+        }
+
+        private static Pulumirpc.CallResponse.Types.ReturnDependencies BuildReturnDependencies(PropertyDependencies dependencies)
+        {
+            var propertyDependencies = new Pulumirpc.CallResponse.Types.ReturnDependencies();
+            propertyDependencies.Urns.AddRange(dependencies.Urns);
+            return propertyDependencies;
+        }
+
+        private static IEnumerable<Pulumirpc.CheckFailure> MappFailures(IEnumerable<CheckFailure> failures)
```
Frassle, line 1112: Map
```suggestion
        private static IEnumerable<Pulumirpc.CheckFailure> MapFailures(IEnumerable<CheckFailure> failures)
```

sdk/Pulumi/Provider/Provider.cs
```diff
@@ -916,14 +966,156 @@ public override async Task<Empty> Delete(Pulumirpc.DeleteRequest request, Server
...
+                var opts = new ComponentResourceOptions()
+                {
+                    Aliases = aliases,
+                    DependsOn = dependsOn,
+                    Protect = request.Protect,
+                    Providers = providers,
+                    Parent = request.Parent != null ? new DependencyResource(request.Parent) : null,
+                };
+
+                var domRequest = new ConstructRequest(request.Name, request.Type, Marshal(request.Inputs), opts);
```
Frassle, line 991: This needs to get the resource monitor endpoint from the request and setup everything for using that.

Frassle (commented):
We'll definitely want an integration test for this.

algompluecker:
I have done some refactoring in order to prepare integration tests in another PR: [PR 227](https://github.com/pulumi/pulumi-dotnet/pull/277)

algompluecker:
I have implemented integration tests now. Please have a look.

sdk/Pulumi.Tests/Resources/CustomTimeoutsTests.cs
```diff
@@ -13,13 +13,13 @@ public void TestDeserialize()
             var rpc = new Pulumirpc.RegisterResourceRequest.Types.CustomTimeouts
             {
                 Create = "1m",
-                Update = "1h3m12.99ms100us",
+                Update = "1h3m12.99ms1us100ns",
```
Frassle, line 16: Good catch, I'd really like this fix pulled into its own PR with a changelog entry. We can merge it very quickly.

sdk/Pulumi.Tests/XUnit/DelegatedLoggerProvider.cs
```diff
@@ -0,0 +1,26 @@
+using System;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+
+namespace DevSpells.WebApi.Testing;
+
+public sealed class DelegatedLoggerProvider : ILoggerProvider
+{
+    private readonly NullLoggerProvider fallback = NullLoggerProvider.Instance;
+    private readonly Func<ILoggerProvider?> providerAccess1;
```
Frassle, line 10:
```suggestion
    private readonly Func<ILoggerProvider?> providerAccess;
```

Zaid-Ajaj:
Hi there @algompluecker thanks a lot for the PR, this is awesome work 🚀 and we are excited to get it over the line. 

There are a few things worth discussing, first is the `ValueWithDependencies`. Why was this necessary to introduce? Not all values have dependencies, only non-plain ones that are outputs or secrets and those are marked as such at the gRPC layer. I think we only want to have dependencies when property values are `OutputReference`.

Second is adding a full integration test, end to end which uses the Pulumi CLI, not just in an isolated unit test within the project. The way to do this adding a new test in `integration_tests` where one directory of that tests is a dotnet console app that is the provider, then another directory is a Pulumi YAML program which uses the built provider binary:
```yaml
name: testingdotnetmlc
runtime: yaml

plugins:
  providers:
    - name: test
      path: ./bin/Debug/net7.0/publish

resources:
  testmlc:
    type: test:index:Test
    properties:
      passwordLength: 20

outputs:
  length: ${testmlc.passwordResult}
```
This would exercise the entire tool chain and we can assert that the e2e behaviour. 

When you are writing a dotnet console app as the provider, make sure that to modify the `AssemblyName` project property so that Pulumi YAML can pick up the provider by name. 
 ```xml
<AssemblyName>pulumi-resource-test</AssemblyName>
```
If you have any questions about this, please let us know and we can help get this in as soon as possible

**Comment by Frassle** [comment]
>  path: ./bin/Debug/net7.0/publish

N.B. I think it will work to point directly at the provider source folder. Shimless should kick in and do the build itself.

**Comment by jkerken** [comment]
@Zaid-Ajaj @Frassle  We encountered a bug where an unkown Input would be deserialized in the provider as an Known Output with a null Value

I added an Integration test in `construct_component_methods_unknown`
Fix is in https://github.com/pulumi/pulumi-dotnet/pull/275/commits/64b04b088c83caecb1286e840677baa584cd1d6e

Please let me know if this a correct (it works for the integration test and our real provider where we discovered this)

**Comment by Frassle** [comment]
> We encountered a bug where an unkown Input would be deserialized in the provider as an Known Output with a null Value

Yeh checked with the other SDKs this looks right to me. Again, if you raise that as its own PR we could merge it ASAP.

**Comment by NatElkins** [comment]
Really looking forward to this feature!!

**Comment by NatElkins** [comment]
Is any help needed to test this?  Happy to pitch in.

**Comment by algompluecker** [comment]
@NatElkins thanks for offering help. Although this PR seems not to be moving, a lot is happening. We are currently implementing the boilerplate and a full C# provider on top of this branch in order to see if everything is working as expected. You might have seen that we have created some separate PRs fixing bugs we have encountered. I am going to finalize this PR very soon, it is mainly about integration tests failing due to some more bugs in the existing code that I need to fix.

**Comment by algompluecker** [comment]
Hi,

I have finalized this PR. Integration tests have been added and are passing.
We have done an internal code review and applied required fixes.

@Frassle @Zaid-Ajaj can you please do a final review?

I can't wait to see this being shipped :)

File: sdk/Pulumi/Provider/Provider.cs [hunk 6]
```diff
@@ -531,34 +633,31 @@ private CancellationTokenSource GetToken(ServerCallContext context)
 
         // Helper to deal with the fact that at the GRPC layer any Struct property might be null. For those we just want to return empty dictionaries at this level.
         // This keeps the PropertyValue.Marshal clean in terms of not handling nulls.
-        private ImmutableDictionary<string, PropertyValue> Marshal(Struct? properties)
+        private ImmutableDictionary<string, PropertyValue> Unmarshal(Struct? properties, IDictionary<string, ISet<Urn>>? inputDependencies = default)
```
Frassle, line 636: I've renamed this in https://github.com/pulumi/pulumi-dotnet/pull/334. You'll want to rebase.

sdk/Pulumi/Provider/ComponentResourceProviderBase.cs
```diff
@@ -0,0 +1,91 @@
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Threading;
+using System.Threading.Tasks;
+using Pulumi.Utilities;
+
+namespace Pulumi.Experimental.Provider;
+
+public class ComponentResourceProviderBase : Provider
```
Frassle, line 10: I think this could just get merged into Provider. There are cases of providers that support Custom and Component resources, there's no need to make components require a different base class for that.

sdk/Pulumi/Provider/PropertyValueSerializer.cs
Frassle, line 1: These serializer changes look like
A) They could go to their own PR
B) Should have tests in sdk/Pulumi.Tests/Provider/PropertyValueTests.cs to match

sdk/Pulumi/Deployment/IDeploymentBuilder.cs
```diff
@@ -0,0 +1,7 @@
+namespace Pulumi;
+
+internal interface IDeploymentBuilder
```
Frassle, line 3: I think its worth pulling this IDeploymentBuilder change into it's own PR and seeing if it makes sense to optionally expose it to non-service tests as well.

sdk/Pulumi/Deployment/IDeploymentInternal.cs
```diff
@@ -10,7 +10,7 @@ internal interface IDeploymentInternal : IDeployment
         string? GetConfig(string fullKey);
         bool IsConfigSecret(string fullKey);
 
-        Stack Stack { get; set; }
+        Stack? Stack { get; set; }
```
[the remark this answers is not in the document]
Zaid-Ajaj: Not with MLCs, I remember having to do the same when I built the PoC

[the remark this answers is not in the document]
Frassle: I'm pretty sure this will lead to issue where Parent isn't explicitly set. That's something I've tried to _and failed so far_ to fix in the engine so we need SDKs to be correct here.

algompluecker:
Hi @Frassle @Zaid-Ajaj I have resolved all the points as discussed in the meetings. Please double check and resolve the remaining open discussions if you agree they are solved.
I think we should be good to merge this now!

Frassle (approved):

Frassle:
I think the test failures are real issues due to the changes in RunInlineAsync. It's not handling exceptions exactly the same as before.

Frassle:
It might be worth trying to do the change to generic <T> in it's own smaller PR.

PR #293, opened 2024-07-07
Introduce type for URN values

algompluecker:


sdk/Pulumi/Provider/UrnValue.cs
```diff
@@ -0,0 +1,11 @@
+using System;
+
+namespace Pulumi.Experimental.Provider;
+
+public record struct UrnValue(string Value)
```
Frassle, line 5:
```suggestion
public record struct Urn(string value)
```

algompluecker: There is already a static class Pulumi.Urn having some helper methods. I have named the new type UrnValue in order to avoid conflicts, but I can rename it if you think this would be better.

Frassle: This is in it's own namespace so I think using `Urn` again should be fine. Long term thinking about moving this out of the experimental namespace we'd probably just merge the two classes into one.

Frassle (approved):
LGTM

PR #296, opened 2024-07-11
[automation] Implement Stack.ImportAsync() for batch importing resources into a stack

Zaid-Ajaj:
Addressing https://github.com/pulumi/pulumi/issues/8237

Following a similar pattern to how I implemented the NodeJS API https://github.com/pulumi/pulumi/pull/16615

This PR extends the automation API to allow batch importing resources into a stack. An example of how this looks like:
```cs
var result = await stack.ImportAsync(new()
{
    Protect = false,
    Resources = new()
    {
        new()
        {
            Type = "random:index/randomPassword:RandomPassword",
            Name = "randomPassword",
            Id = "supersecret"
        }
    }
});
```

Also implements an enum converter for `UpdateKind` since JSON deserialization was failing when encountering update kind `resource-import`, this is now fixed.

### DIFF
--- a/sdk/Pulumi.Automation/ImportResource.cs
+++ b/sdk/Pulumi.Automation/ImportResource.cs
@@ -1,0 +1,33 @@
+namespace Pulumi.Automation;
+
+using System.Collections.Generic;
+
+public class ImportResource
+{
+    /// <summary>
+    /// The name of the resource to import
+    /// </summary>
+    public string? Name { get; init; }
+    /// <summary>
+    /// The type of the resource to import
+    /// </summary>
+    public string? Type { get; init; }
+    /// <summary>
+    /// The ID of the resource to import. The format of the ID is specific to the resource type
+    /// </summary>
+    public string? Id { get; init; }
+
+    public string? Parent { get; init; }
+
+    public string? Provider { get; init; }
+
+    public string? Version { get; init; }
+
+    public string? LogicalName { get; init; }
+
+    public List<string>? Properties { get; init; }
+
+    public bool? Component { get; init; }
+
+    public bool? Remote { get; init; }
+}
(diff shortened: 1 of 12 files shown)


sdk/Pulumi.Automation/ImportResource.cs
```diff
@@ -0,0 +1,33 @@
+namespace Pulumi.Automation;
+
+using System.Collections.Generic;
+
+public class ImportResource
```
Frassle, line 5: It is very sad these need to be kept in sync manually across all the SDKs. We really need automation V2 to make this work scalable :(

sdk/Pulumi.Automation/WorkspaceStack.cs
```diff
@@ -636,6 +638,105 @@ void OnPreviewEvent(EngineEvent @event)
...
+
+                if (options.Converter != null)
+                {
+                    // if the user specifies a converter, pass it to `--from <converter>` argument of import.
+                    args.Add("--from");
+                    args.Add(options.Converter);
+                    if (options.ConverterArgs?.Any() == true)
+                    {
+                        // pass any additional arguments to the converter
+                        args.AddRange(options.ConverterArgs);
```
Frassle, line 713: Should probably add a "--" arg here to split the converter args from the args to import

Zaid-Ajaj: Done ✅

sdk/Pulumi.Automation/ImportResource.cs
Zaid-Ajaj: I worked on a PoC a while back where you could specify types and modules in `.proto` files and showed that it could actually work to generate types but it would only generate about 80-90% of things, not every type we have.

Frassle (approved):

PR #298, opened 2024-07-17
Add support for deserializing output values

### LINKED ISSUE #297: Transforms aren't deserializing output values
Consider this program:

```c#
using Pulumi;
using Pulumi.Aws.S3;
using System;
using System.Collections.Immutable;

await Deployment.RunAsync<MyStack>();

class MyStack : Stack
{
    public MyStack() : base(new StackOptions
    {
        ResourceTransforms =
        {
            async (args, _) => {
                if (args.Type == "aws:s3/bucket:Bucket")
                {
                    if (args.Args.TryGetValue("tags", out var tagsValue) && tagsValue is ImmutableDictionary<string, object> tagsDictionary)
                    {
                        foreach (var (key, value) in tagsDictionary)
                        {
                            Console.WriteLine($"{key}: {value}");
                        }
                    }
                }
                return null;
            },
        }
    })
    {
        var bucket = new Bucket("my-bucket", new BucketArgs
        {
            Tags = Output.CreateSecret(new InputMap<string>
            {
                { "env", "dev" },
            }.ToOutput()),
        });
    }
}
```

## Actual

The value of the `"tags"` key in `args.Args` is basically:

```c#
ImmutableDictionary.Create<string, object>()
    .Add("4dabf18193072939515e22adb298388d", "d0e6a833031e9bbcd3f4e8bde6ca49a4")
    .Add("secret", true)
    .Add("value", ImmutableDictionary.Create<string, object>().Add("env", "dev"));
```

## Expected

It should be `Output<ImmutableDictionary<string, object>>` with the secret bit set to true internally.

justinvp:
This change adds support for deserializing output values and ensures they work for transforms. This is necessary as the engine will send output values to transforms, so we need to handle deserializing them before calling the transform and then properly serializing them when returning the results.

Fixes #297

### DIFF
--- a/sdk/Pulumi/Deployment/Deployment_Prepare.cs
+++ b/sdk/Pulumi/Deployment/Deployment_Prepare.cs
@@ -4,10 +4,10 @@ using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
-using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
 using Google.Protobuf.WellKnownTypes;
+using Pulumi.Serialization;
 
 namespace Pulumi
 {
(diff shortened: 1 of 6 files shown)


sdk/Pulumi/Deployment/Deployment_Prepare.cs
```diff
@@ -166,60 +166,10 @@ private static async Task<Pulumirpc.Callback> AllocateTransform(Callbacks callba
...
-                                .GetMethod("FromResult")!
-                                .MakeGenericMethod(outputDataType)!;
-
-                        var outputObject = createOutputMethod.Invoke(new[] { fromResultMethod.Invoke(null, new[] { typedOutputData }) });
-                        props.Add(kv.Key, outputObject);
-                    }
+                    // The data from the engine will include output values, which will be deserialized
+                    // into instances of `Output<T>`, which will track whether the value is known,
+                    // is secret, and dependencies. We can ignore those values on outputData.
+                    props.Add(kv.Key, outputData.Value);
```
Frassle, line 172: Doesn't this still need to check the OutputData for known/secretness? Doesn't look like Deserialise is returning Output<T> for those, just for actual output wire objects.

justinvp: I don't think so. Shouldn't everything be coming across from the engine to transforms as output values? In which case, those will already have the known/secretness/dependency tracking.

Frassle: For transforms yes, but isn't this used for other deserialisation for invokes/resources?

Frassle: Ah no ignore me, this is specific to transforms. Should be fine then.

Frassle (approved):

PR #299, opened 2024-07-17
Update .xml doc files

justinvp:
Noticed some changes to these files during local development that hadn't been committed. This change does that.


tgummerer (approved):
Out of curiosity, which command regenerates them?  Also should we add a CI check making sure they are up to date?

justinvp:
> Out of curiosity, which command regenerates them? Also should we add a CI check making sure they are up to date?

Just doing a build regenerates them. It indeed would be nice to add a CI check. This annoyingly happens a lot.

PR #306, opened 2024-07-24
Fix unknown inputs deserialization

jkerken:
Fix extracted from https://github.com/pulumi/pulumi-dotnet/pull/275

I also found the same bug for unkown secrets inputs. Which is also fixed.

I did not copy the intgeration test over since that was done using a c# component provider
But it is present in  the other pr, I added some unit tests to cover these edge cases though.

### DIFF
--- a/.changes/unreleased/Bug Fixes-306.yaml
+++ b/.changes/unreleased/Bug Fixes-306.yaml
@@ -1,0 +1,6 @@
+component: sdk
+kind: Bug Fixes
+body: Fix unkown inputs deserialization
+time: 2024-07-24T14:53:11.4857503+02:00
+custom:
+    PR: "306"
--- a/.changes/unreleased/Bug Fixes-306.yaml
+++ b/.changes/unreleased/Bug Fixes-306.yaml
@@ -1,6 +1,6 @@
 component: sdk
 kind: Bug Fixes
-body: Fix unkown inputs deserialization
+body: Fix unknown inputs deserialization
 time: 2024-07-24T14:53:11.4857503+02:00
 custom:
     PR: "306"
--- a/sdk/Pulumi.Tests/Provider/PropertyValueTests.cs
+++ b/sdk/Pulumi.Tests/Provider/PropertyValueTests.cs
@@ -282,6 +282,22 @@ public class PropertyValueTests
     }
 
     [Fact]
+    public async Task DeserializingUnknownOutputWorks()
+    {
+        var serializer = CreateSerializer();
+        var output = new PropertyValue(new OutputReference(
+            value: PropertyValue.Computed,
+            dependencies: ImmutableArray<Urn>.Empty));
+
+        var deserialized = await serializer.Deserialize<Input<string>>(output);
+        var deserializedOutput = deserialized.ToOutput();
+        var data = await deserializedOutput.DataTask;
+        Assert.Null(data.Value);
+        Assert.False(data.IsSecret);
+        Assert.False(data.IsKnown);
+    }
+
+    [Fact]
     public async Task DeserializingWrappedOutputSecretWorks()
     {
         var serializer = CreateSerializer();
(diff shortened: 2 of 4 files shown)


.changes/unreleased/Bug Fixes-304.yaml
```diff
@@ -1,6 +1,6 @@
 component: sdk
 kind: Bug Fixes
-body: Fix race condition in GrpcMonitor's GrpcChannel management
```
Frassle, line 3: Can leave this file as is

sdk/Pulumi.Tests/Provider/PropertyValueTests.cs
```diff
@@ -299,6 +315,24 @@ public async Task DeserializingWrappedOutputSecretWorks()
         Assert.True(data.IsKnown);
     }
 
+    [Fact]
+    public async Task DeserializingWrappedUnknownOutputSecretWorks()
+    {
+        var serializer = CreateSerializer();
+        // secretOutput = Secret(Output(Unknown))
```
Frassle, line 322: ?

sdk/Pulumi.Tests/Provider/PropertyValueTests.cs
```diff
@@ -299,6 +315,24 @@ public async Task DeserializingWrappedOutputSecretWorks()
...
+        var secretOutput = new PropertyValue(
+            new PropertyValue(new OutputReference(
+                value:  PropertyValue.Computed,
+                dependencies: ImmutableArray<Urn>.Empty)));
```
Frassle, line 326: Shouldn't "secret" be on the OutputReference object?

.changes/unreleased/Bug Fixes-306.yaml
```diff
@@ -0,0 +1,6 @@
+component: sdk
+kind: Bug Fixes
+body: Fix unkown inputs deserialization
```
Frassle, line 3:
```suggestion
body: Fix unknown inputs deserialization
```

sdk/Pulumi.Tests/Provider/PropertyValueTests.cs
Frassle: Ah no, this just needs to wrap in the double PropertyValue as it's doing.

jkerken: In deserialize there are both cases covered.

- Secret(Output(...)) being deserialized into Input<T>
- Output(Secret(...)) being deserialized into Input<T>

There is also already: 

- DeserializingWrappedSecretInOutputWorks which deserializes Output(Secret("Hello"))
- DeserializingWrappedOutputSecretWorks which deserializes Secret(Output("Hello"))

I just assumed that both value can be unknown

.changes/unreleased/Bug Fixes-306.yaml
jkerken: Fixed

sdk/Pulumi.Tests/Provider/PropertyValueTests.cs
Frassle: Yeh either case could be unknown.

sdk/Pulumi.Tests/Provider/PropertyValueTests.cs
Frassle, line 322:
```suggestion

```

sdk/Pulumi.Tests/Provider/PropertyValueTests.cs
```diff
@@ -317,6 +351,25 @@ public async Task DeserializingWrappedSecretInOutputWorks()
         Assert.True(data.IsKnown);
     }
 
+
+    [Fact]
+    public async Task DeserializingWrappedUnknownSecretInOutputWorks()
+    {
+        var serializer = CreateSerializer();
+        // secretOutput = Output(Secret(Unknown))
```
Frassle, line 359: I think these comments caused me more confusion than understanding, I'd just take them out. If you want a description here put it in English "secretOutput is Unknown wrapped in a Secret wrapped in an Output"

Zaid-Ajaj (approved):

PR #309, opened 2024-07-29
Fix handling of input properties with backing fields

jkerken:
Fixes https://github.com/pulumi/pulumi-dotnet/issues/308

I also limited the properties that are tried to be deserialized to not include non instance fields.
The generated args classes all have a static Empty Property that threw an error otherwise.

### DIFF
--- a/sdk/Pulumi.Tests/Provider/PropertyValueTests.cs
+++ b/sdk/Pulumi.Tests/Provider/PropertyValueTests.cs
@@ -68,8 +68,10 @@ public class PropertyValueTests
             Pair("password", new PropertyValue(new PropertyValue("PW"))));
 
         var basicArgs = await serializer.Deserialize<SecretArgs>(data);
-        var pw = await basicArgs.Password.ToOutput().GetValueAsync("");
-        Assert.Equal("PW", pw);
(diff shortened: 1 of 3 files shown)


sdk/Pulumi.Tests/Provider/PropertyValueTests.cs
```diff
@@ -45,6 +45,33 @@ public async Task DeserializingBasicArgsWorks()
...
+    [Fact]
+    public async Task DeserializingSecretArgsWorks()
+    {
+        var serializer = CreateSerializer();
+        var data = Object(
+            Pair("password", new PropertyValue(new PropertyValue("PW"))));
+
+        var basicArgs = await serializer.Deserialize<SecretArgs>(data);
+        var pw = await basicArgs.Password.ToOutput().GetValueAsync("");
+        Assert.Equal("PW", pw);
```
Frassle, line 72: Can this assert that Password is secret as well?

jkerken: done

Zaid-Ajaj (approved):
Looks great!

jkerken:
@Frassle Can we merge this?

PR #313, opened 2024-08-06
Updated SDK docs on output

pierskarsenbarg:
The previous version was technically correct in that you shouldn't create resources within an apply function but was incorrect that it is not allowed

### DIFF
--- a/sdk/Pulumi/Core/Output.cs
+++ b/sdk/Pulumi/Core/Output.cs
@@ -496,7 +496,7 @@ namespace Pulumi
         /// name="func"/>. The result remains an <see cref="Output{T}"/> so that dependent resources
         /// can be properly tracked.
         /// <para/>
-        /// <paramref name="func"/> is not allowed to make resources.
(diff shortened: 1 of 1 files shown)


pierskarsenbarg:
See also pulumi/pulumi#16885

sdk/Pulumi/Core/Output.cs
```diff
@@ -496,7 +496,7 @@ public Output<U> Apply<U>(Func<T, Input<U>?> func)
         /// name="func"/>. The result remains an <see cref="Output{T}"/> so that dependent resources
         /// can be properly tracked.
         /// <para/>
-        /// <paramref name="func"/> is not allowed to make resources.
+        /// <paramref name="func"/> should not be used to create resources unless necessary. Creating resources within 'func' can cause unknown behaviour.
```
tgummerer, line 499:
```suggestion
        /// <paramref name="func"/> should not be used to create resources unless necessary as `func` may not be run during some program executions.
```

Let's make it the same as in https://github.com/pulumi/pulumi/pull/16885.

tgummerer (approved):
Sorry about the delay reviewing this.

PR #329, opened 2024-08-30
add Makefile to build the language runtime

tgummerer:
Add a simple makefile to build and install the language runtime for dotnet.  We have a homegrown build system in this repo, but that doesn't even return the errors for failed Go builds.  Rather than trying to fix that, it's better to try and converge the build systems for all our projects, as all others are using Make.

The `dotnet run` commands will keep working for those that prefer them.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,9 @@
+GO                          := go
+
+install::
+	cd pulumi-language-dotnet && ${GO} install ./...
+
+build::
+	cd pulumi-language-dotnet && ${GO} build .
+
+.PHONY: install build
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -0,0 +1,9 @@
+GO                          := go
```
lunaris, line 1: This seems like some aggressive spacing :joy: Unless this is Make-induced-tab-requiredness?

lunaris (approved):

Makefile
tgummerer: lol that's copy pasta spacing.  I'll change that.

PR #332, opened 2024-09-04
implement debugging support for dotnet

### LINKED ISSUE #333: Debugger Support: .NET
Add debugging support to the .NET language host.

tgummerer:
For dotnet to run with a debugger, we don't need to run the binary under a debugger, but rather the IDEs can just attach to the program. This needs some support from the SDK, so we wait until the debugger is actually attached.

Note that this does not seem to work when just running the program using `dotnet run .`, but rather we need to build and publish the program first, and only then we can attach the debugger to it.  (It might just be that running the program under the debugger is horribly slow, and thus using `dotnet run` takes more time than I had patience for, but that wouldn't be a good user experience either way).

I'm not sure how to write an integration test for this.  For Go the debugger starts a tcp server we can communicate with, but not sure what attaching to a PID means in that context/how exactly to do that.  I did try it out manually (which was also difficult, since including the local SDK doesn't seem to work with `dotnet build; dotnet <path-to.dll>`, so I ended up just copying the SDK changes to a program :grimacing:)

Fixes #333


sdk/Pulumi/Deployment/Deployment_Run.cs
```diff
@@ -18,6 +18,15 @@ public partial class Deployment
         public static Task<int> RunAsync(Action action)
             => RunAsync(() =>
             {
+                var value = System.Environment.GetEnvironmentVariable("PULUMI_ATTACH_DEBUGGER");
+                if (value != null && value == "true")
+                {
+                    while (!System.Diagnostics.Debugger.IsAttached)
+                    {
+                        // keep waiting until the debugger is attached
+                        System.Threading.Tasks.Task.Delay(1);
```
Frassle, line 27: Need to await this

pulumi-language-dotnet/main.go
```diff
@@ -547,8 +559,49 @@ func (w *logWriter) LogToUser(val string) (int, error) {
 	return len(val), nil
 }
 
+func (host *dotnetLanguageHost) buildDll() error {
+	// If we are running from source, we need to build the project.
+	// Run the `dotnet build` command.  Importantly, report the output of this to the user
+	// (ephemerally) as it is happening so they're aware of what's going on and can see the progress
+	// of things.
+	args := []string{"build", "-nologo"}
```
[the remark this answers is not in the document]
tgummerer: Hmm is that not the default profile?  At least when I just run `dotnet build`, the artifacts end up in a `bin/Debug/...` directory.

pulumi-language-dotnet/main.go
```diff
@@ -547,8 +559,49 @@ func (w *logWriter) LogToUser(val string) (int, error) {
...
+
+	cmd := exec.Command(host.exec, args...)
+	err := cmd.Run()
+	if err != nil {
+		return errors.Wrapf(err, "failed to build project: %v", err)
+	}
+
+	args = []string{"publish", "-o", "bin/publish"}
```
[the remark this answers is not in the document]
tgummerer: I thought it was needed, but looks like just build is enough after all.  Removed the publish step.

sdk/Pulumi/Deployment/Deployment_Run.cs
tgummerer: We're not in an `async` context here, so I switched it out for `System.Threading.Thread.Sleep(1);`.  Or is there a strong reason why we should be using `Delay`?

Frassle: Delay is better if in an async context. Sleep works fine here.

pulumi-language-dotnet/main.go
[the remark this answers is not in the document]
Frassle: It should be the default for most projects. Just gonna be one of those annoying edge case things that'll be grim to fix and we'll have to hope doesn't come up.

pulumi-language-dotnet/main.go
```diff
@@ -547,8 +559,42 @@ func (w *logWriter) LogToUser(val string) (int, error) {
...
+		return errors.Wrapf(err, "failed to build project: %v", err)
+	}
+
+	err = filepath.WalkDir(".", func(path string, d os.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if name, ok := strings.CutSuffix(d.Name(), ".csproj"); ok {
+			host.binary = filepath.Join("bin", "pulumi-debugging", name+".dll")
```
Frassle, line 581: This probably shouldn't change global state. A language plugin can be used for multiple things in parallel and this is gonna confuse things.
Honestly "binary" shouldn't even be a global var to start with, the host should be re-parsing it from options on each run.

pulumi-language-dotnet/main.go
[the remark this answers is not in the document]
tgummerer: :crossed_fingers:

pulumi-language-dotnet/main.go
tgummerer: Changed that now.  I haven't fixed the global `host.binary` though, that's probably something to look at in a separate PR.

PR #338, opened 2024-09-06
Add mapFailures helper to provider

Frassle:
Idea taken from https://github.com/pulumi/pulumi-dotnet/pull/275, where there were even more cases of translating CheckFailure objects.

But this is a small clean change to pull in by itself.

### DIFF
--- a/sdk/Pulumi/Provider/Provider.cs
+++ b/sdk/Pulumi/Provider/Provider.cs
@@ -540,6 +540,21 @@ namespace Pulumi.Experimental.Provider
             return PropertyValue.Unmarshal(properties);
         }
 
+        // Helper to marshal CheckFailures from the domain to the GRPC layer.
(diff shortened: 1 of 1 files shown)


sdk/Pulumi/Provider/Provider.cs
```diff
@@ -540,6 +540,21 @@ private ImmutableDictionary<string, PropertyValue> Unmarshal(Struct? properties)
             return PropertyValue.Unmarshal(properties);
         }
 
+        // Helper to marshal CheckFailures from the domain to the GRPC layer.
+        private IEnumerable<Pulumirpc.CheckFailure> mapFailures(IEnumerable<CheckFailure>? failures)
```
Zaid-Ajaj, line 544: Can you rename this to `MapFailures` instead? (as per C# standards)

Frassle: I switch languages too much 😆

Zaid-Ajaj (approved):

PR #343, opened 2024-09-13
Fix debugger support

### LINKED ISSUE #344: Running `pulumi preview --attach-debugger` for a .NET program doesn't wait for the debugger to be attached
Two issues:

1. Only one overload of `RunAsync` is waiting for the debugger to be attached, so if you are calling any of the other overloads (like the ones used in project templates) it won't wait for the debugger to be attached.

2. If you do attach the process, breakpoints don't appear to work. It looks like this is because it's still running `dotnet run .` instead of the intended `dotnet <name>.dll` when `--attach-debugger` is specified. This is due to a variable being shadowed unintentionally in the lang host.

justinvp:
Two fixes for debugger support:

1. Fix unintentional shadowing of the `binaryPath` variable in the langhost's `Run` method. It's currently being shadowed, which means we won't run `dotnet <name>.dll` needed to be able to attach a debugger. (This "regressed" when addressing PR feedback in #332).

2. Fix where we're waiting for the debugger to be attached in the SDK. Right now, we're only waiting in a single overload of `Deployment.RunAsync`. If your Pulumi program is calling another overload, it won't wait for the debugger to attach. This change moves the logic to a centralized place that all `RunAsync` overloads call through to. It also converts the logic back into be async by awaiting `Task.Delay` with a one second delay, since it can now be async in this new location.

Fixes #344

### DIFF
--- a/sdk/Pulumi/Deployment/Deployment_Run.cs
+++ b/sdk/Pulumi/Deployment/Deployment_Run.cs
@@ -337,7 +337,7 @@ namespace Pulumi
                 while (!System.Diagnostics.Debugger.IsAttached)
                 {
                     // keep waiting until the debugger is attached
-                    await Task.Delay(1000).ConfigureAwait(false);
+                    await Task.Delay(100).ConfigureAwait(false);
                 }
             }
 
--- a/sdk/Pulumi/Deployment/Deployment_Run.cs
+++ b/sdk/Pulumi/Deployment/Deployment_Run.cs
@@ -18,15 +18,6 @@ namespace Pulumi
         public static Task<int> RunAsync(Action action)
             => RunAsync(() =>
             {
-                var value = System.Environment.GetEnvironmentVariable("PULUMI_ATTACH_DEBUGGER");
-                if (value != null && value == "true")
-                {
-                    while (!System.Diagnostics.Debugger.IsAttached)
-                    {
-                        // keep waiting until the debugger is attached
-                        System.Threading.Thread.Sleep(1);
-                    }
-                }
                 action();
                 return ImmutableDictionary<string, object?>.Empty;
             });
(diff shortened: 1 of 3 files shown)


Frassle (approved):

sdk/Pulumi/Deployment/Deployment_Run.cs
```diff
@@ -341,6 +332,15 @@ internal static async Task<int> CreateRunnerAndRunAsync(
             Func<Deployment> deploymentFactory,
             Func<IRunner, Task<int>> runAsync)
         {
+            if (Environment.GetEnvironmentVariable("PULUMI_ATTACH_DEBUGGER") == "true")
+            {
+                while (!System.Diagnostics.Debugger.IsAttached)
+                {
+                    // keep waiting until the debugger is attached
+                    await Task.Delay(1000).ConfigureAwait(false);
```
mikhailshilkov, line 340: Nice change from `Thread.Sleep(1)`: both converting it to async and waiting longer than 1 ms. I don't think 1s is noticeable that much, but should we go for 100 ms instead to make sure the experience is as best as possible?

mikhailshilkov (approved):
LGTM. Missing tests are a bit sad but I also see that the original PR didn't have any, which is double sad. I don't think we should block on that here, but I'd love to see a follow-up task to add them retroactively.

sdk/Pulumi/Deployment/Deployment_Run.cs
justinvp: I'd tried 500ms and 1000ms and didn't see any noticeable differences in the experience from my machine, but went ahead and reduced to 100ms to make sure things are fast when attaching.

justinvp:
> Missing tests are a bit sad but I also see that the original PR didn't have any, which is double sad.

Agreed. We've been thinking about the best way to test this, but didn't want to block on it given the timing. (Though, tests likely would have avoided the need for this PR). I've opened https://github.com/pulumi/pulumi-dotnet/issues/345 to track and included an idea on how we might test it.

Issue #345, opened 2024-09-13
Add test coverage for debugger support


justinvp:
We need to add some testing around the debugger support in the .NET langhost/SDK. Perhaps we can do something along the lines of the integration test we added for Go in https://github.com/pulumi/pulumi/pull/17102?

Maybe attaching with https://github.com/Samsung/netcoredbg?

PR #349, opened 2024-09-17
add integration test for debugging dotnet

tgummerer:
We can do this by attaching netcoredbg to the running process, and sending it a command, so the process realizes a debugger is attach. We're using the thread-info command here, since that's one that works after a bit of experimentation.

Fixes https://github.com/pulumi/pulumi-dotnet/issues/345


integration_tests/integration_dotnet_test.go
```diff
@@ -535,42 +545,127 @@ func TestDeletedWith(t *testing.T) {
 func TestProviderCall(t *testing.T) {
```
tgummerer, line 545: Looks like some stuff here wasn't properly gofmt'd, so this diff is unfortunately a bit noisier than it should be.

integration_tests/integration_dotnet_test.go
```diff
@@ -535,42 +545,127 @@ func TestDeletedWith(t *testing.T) {
...
+			return nil, nil
+		}
+		return nil, fmt.Errorf("expected to be able to open event log file %s: %w",
+			logfile, err)
+	}
+
+	defer contract.IgnoreClose(eventsFile)
+
+	//	f, _ := ioutil.ReadAll(eventsFile)
+	//	fmt.Println("eventsfile", string(f))
```
justinvp, line 602: Do we need to keep these?

tgummerer: Whoops, no, this was just for debugging and I forgot to remove it.

PR #360, opened 2024-10-02
always run all integration tests

tgummerer:
Currently the CI workflow in this repo runs each integration test individually.  This makes it very easy to forget to add a test to the list, that's then never run in CI, even though it should be.

Fix this by introducing a new Makefile target, to run all integration tests, and running that in CI instead of spelling each test out individually, which is very brittle.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -110,7 +110,7 @@ jobs:
         run: |
           scoop install netcoredbg
       - name: Install netcoredbg (Linux)
-        if: matrix.os == 'ubuntu-latest'
+        if: matrix.os == 'macos-latest'
         run: |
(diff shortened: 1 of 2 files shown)


lunaris (approved):
I'm guessing it's not safe to run them in parallel?

tgummerer:
I'm still trying to figure out why a bunch of tests are failing.  Them running in parallel was my first suspicion, but you might have to bear with me a little longer to try and figure this out.

tgummerer:
> I'm guessing it's not safe to run them in parallel?

No, we should probably follow up on that and try to run them in parallel more, but they already don't run in parallel, and I just want to make sure we run all tests (I recently added one that was never run in CI, and therefore obviously it was broken, though it did work locally). I think fixing that footgun is momre important for now, and then we can add other improvements on top.

PR #361, opened 2024-10-03
WIP: allow dotnet to emit better errors for component construct

tgummerer:
Introduce a new InputPropertiesError that allows users to return more error details from a component construct.

### DIFF
--- a/sdk/Pulumi/Exceptions/InputPropertiesException.cs
+++ b/sdk/Pulumi/Exceptions/InputPropertiesException.cs
@@ -1,0 +1,51 @@
+// Copyright 2016-2024, Pulumi Corporation
+
+using System;
+using System.Collections.Generic;
+using Grpc.Core;
+using Pulumirpc;
+
+namespace Pulumi
+{
+	public class PropertyError {
+		public string PropertyPath { get; set; }
+		public string Reason { get; set; }
+
+		public PropertyError(string propertyPath, string reason) {
+			PropertyPath = propertyPath;
+			Reason = reason;
+		}
+	}
+
+	public class InputPropertiesException : RpcException
+	{
+		public InputPropertiesException(String message, IList<PropertyError> propertyErrors)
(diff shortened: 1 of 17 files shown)


sdk/Pulumi/Exceptions/InputPropertiesException.cs
```diff
@@ -0,0 +1,51 @@
...
+				Code = (int)StatusCode.Unknown,
+				Message = "Bad request",
+				Details = { Google.Protobuf.WellKnownTypes.Any.Pack(errorDetails) }
+			};
+
+//			throw new RpcException(new Grpc.Core.Status(StatusCode.InvalidArgument, "wtf"), "failing for a reason");
+
+			// var details = Google.Protobuf.WellKnownTypes.Any.Pack(status);
+			var metadata = new Metadata();
+			metadata.Add("grpc-status-details-bin", Google.Protobuf.MessageExtensions.ToByteArray(status));
```
tgummerer, line 46: This is how I thought the status details need to be added to the exception, but for some reason it doesn't seem to add the details.  
E.g. [here](https://github.com/grpc/grpc-dotnet/blob/490894cdedb34b43c32dc576d5745c41223764b6/src/Grpc.StatusProto/MetadataExtensions.cs#L68) (this one is slightly different since it packs it into an `Any` type bofere or [here](https://github.com/dapr-sandbox/components-dotnet-sdk/blob/1b22d21f87261ea0be5d214bad69a262d7d62fdc/src/Dapr.PluggableComponents/Components/StateStore/ETagErrors.cs#L62) it seems to be done in the same way, but neither of those ways seem to make it work for me.  Not sure if I'm doing something wrong somewhere here :thinking:

justinvp: Note for myself: https://github.com/grpc/grpc-dotnet/tree/master/src/Grpc.StatusProto#server-side seems useful

Issue #203, opened 2023-11-07
Make `Pulumi.RunException` assertable


JasonWhall:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Currently when writing Unit tests for a stack that does `Pulumi.Log.Error("")`. The resulting stack will throw a `Pulumi.RunException` in the deployment runner. As this is an internal only type currently, we are unable to assert to check this error has been thrown. The workaround currently is to assert on the `Exception` type, but isn't accurate to the type of exception thrown. Below shows an example of a test that would check for this:

```csharp
using FluentAssertions;
using NSubstitute;
using Pulumi;
using Pulumi.Testing;

public class UnitTest1
{

    [Fact]
    public async Task Stack_LoggingError_ShouldThrowExpectedException()
    {
        // Arrange
        var mocks = Substitute.For<IMocks>();
        var options = new TestOptions { IsPreview = false };
        static void act()
        {
            Log.Error("Message 1");
        };

        var sut = () => Deployment.TestAsync(mocks, options, act);

        // Act & Assert
        await sut.Should().ThrowAsync<Exception>()
            .WithMessage("Message 1");

        // Desired
        /*
            await sut.Should().ThrowAsync<Pulumi.RunException>()
                .WithMessage("Message 1");
        */
    }
}
```

Could the `RunException` be made public so we can assert on this? 

### Affected area/feature

Pulumi Core

justinvp:
Thanks for the suggestion, @JasonWhall. I don't recall if we had a good reason for making `RunException` internal, other than not intending it to be thrown by users. But I understand why it'd be nice to write a test like this.

PR #364, opened 2024-10-16
feat(#203): RunException access modifier set to public. Docs updated

mmisztal1980:
As per discussion in #203 . This is a simple improvement in order to make the exception assertable by testing frameworks such as `FluentAssertions` using:

```csharp
var action = () => 
{
  // (...) // provision failing pulumi resources here
};

action.Should().Throw<Pulumi.RunException>();
```

This will significantly improve testing scenarios where exceptions are being thrown, as everything gets wrapped in an internal `Pulumi.RunException` while testing:

```csharp
// Disassembled code
internal static async Task<(ImmutableArray<Resource> Resources, Exception? Exception)> TryTestAsync(
            IMocks mocks, Func<IRunner, Task<int>> runAsync, TestOptions? options = null)
        {
            var engine = new MockEngine();
            var monitor = new MockMonitor(mocks);
            await CreateRunnerAndRunAsync(() => new Deployment(engine, monitor, options), runAsync).ConfigureAwait(false);
            Exception? err = engine.Errors.Count switch
            {
                1 => new RunException(engine.Errors.Single()), // <-- wrapping takes place here
                var v when v > 1 => new AggregateException(engine.Errors.Select(e => new RunException(e))),
                _ => null
            };
            return (Resources: monitor.Resources.ToImmutableArray(), Exception: err);
        }
```

### DIFF
--- a/.changes/unreleased/Improvements-364.yaml
+++ b/.changes/unreleased/Improvements-364.yaml
@@ -1,0 +1,6 @@
+component: sdk
+kind: Improvements
+body: Pulumi.RunException is public
+time: 2024-10-16T12:40:31.206421+02:00
+custom:
+    PR: "364"
--- a/.changes/unreleased/Improvements-364.yaml
+++ b/.changes/unreleased/Improvements-364.yaml
@@ -1,6 +1,6 @@
 component: sdk
 kind: Improvements
-body: Pulumi.RunException is public
+body: Make Pulumi.RunException public
 time: 2024-10-16T12:40:31.206421+02:00
 custom:
     PR: "364"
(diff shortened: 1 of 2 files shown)


tgummerer (commented):
Thanks for the PR.  I'll let someone with more context approve and merge this, but do have a small suggestion for a changelog improvement (see also https://github.com/pulumi/pulumi/blob/master/CONTRIBUTING.md#changelog-messages for how we write our changelog messages)

mmisztal1980:
@tgummerer when do You think someone will get the chance to look at this? :)

Zaid-Ajaj:
Hi there @mmisztal1980 thanks a lot for the contribution 🙏 we discussed this with the team and we thought it made sense to make the exception public.

Zaid-Ajaj (approved):

mmisztal1980:
@Zaid-Ajaj awesome!
When is the next package release due to happen?

Zaid-Ajaj:
> When is the next package release due to happen?

@mmisztal1980  Not planned at the moment, but we publish dev sdks from commits merged into main so you will be able to use the an alpha package containing your changes

Issue #367, opened 2024-10-29
Types with nested InputList fail to serialize with JsonSerializer.SerializeAsync


mikhailshilkov:
### What happened?

The following code

```cs
using Pulumi;
using System.Collections.Generic;

return await Pulumi.Deployment.RunAsync(() =>
{
    var inputA = new A { B = [123] };
    var willSerialize = Output.JsonSerialize(Output.Create(inputA));

    var inputListA = new InputList<A> { inputA };
    var willNOTSerialize = Output.JsonSerialize(Output.Create(inputListA));

    return new Dictionary<string, object?>();
});

record A 
{
    public required InputList<int> B { get; set; }
}
```

fails with

```
System.NotSupportedException: The type 'Pulumi.InputList`1[System.Int32]' can only be serialized using async serialization methods. The unsupported member type is located on type 'Pulumi.Input`1[A]'. Path: $.
       at void System.Text.Json.ThrowHelper.ThrowNotSupportedException(ref WriteStack state, NotSupportedException ex)
       at bool System.Text.Json.Serialization.JsonConverter<T>.WriteCore(Utf8JsonWriter writer, in T value, JsonSerializerOptions options, ref WriteStack state)
       at async Task System.Text.Json.Serialization.Metadata.JsonTypeInfo<T>.SerializeAsync(Stream utf8Json, T rootValue, CancellationToken cancellationToken, object rootValueBoxed) x 3
       at async Task<OutputData<string>> Pulumi.Output+<>c__DisplayClass13_0<T>.<JsonSerialize>g__GetData|0(?)+GetData(?) ---> System.NotSupportedException: The type 'Pulumi.InputList`1[System.Int32]' can only be serialized using async serialization methods.
       at void System.Text.Json.ThrowHelper.ThrowNotSupportedException_TypeRequiresAsyncSerialization(Type propertyType)
       at bool System.Text.Json.Serialization.Converters.IAsyncEnumerableOfTConverter<TAsyncEnumerable, TElement>.OnTryWrite(Utf8JsonWriter writer, TAsyncEnumerable value, JsonSerializerOptions options, ref WriteStack state)
       at bool System.Text.Json.Serialization.JsonConverter<T>.TryWrite(Utf8JsonWriter writer, in T value, JsonSerializerOptions options, ref WriteStack state)
[...]

PR #371, opened 2024-10-30
Generalize InputList<> serialization to maps and other Input<> subtypes

mikhailshilkov:
#368 fixed `OutputJsonConverter` for `InputList<T>` but it still fails for `InputMap<T>`, so any input maps nested in a serialized data structure causes serialization to fail. Reported by a customer trying to serialize Azure Native dashboard data types.

The fix expands `OutputJsonConverter` to support any type that inherits from `Input<T>`, which includes both `InputList` and `InputMap`. It piggybacks on the existing `InputJsonConverterInner` implementation by passing the right generic argument to it.

Fixes https://github.com/pulumi/pulumi-dotnet/issues/367

Generalizes #368

### DIFF
--- a/.changes/unreleased/bug-fixes-368.yaml
+++ b/.changes/unreleased/bug-fixes-368.yaml
@@ -1,6 +1,6 @@
 component: sdk
 kind: bug-fixes
-body: Support input lists in JsonSerializer.SerializeAsync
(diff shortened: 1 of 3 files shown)


justinvp (approved):
LGTM

Issue #370, opened 2024-10-30
Allow customization of JsonSerializationOptions


rosieks:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Right now to deserialize configuration Pulumi use [default settings for JsonSerializer](https://github.com/pulumi/pulumi-dotnet/blob/2aa84b517a894d432fcb1200531633ed9aee0b9b/sdk/Pulumi/Config.cs#L163). That means that if someone wants to use to use different configuration then it's forced to use eg `JsonPropertyName` and others tools on each class which makes code harder to read. It would be great if there was an option to override it and provide custom `JsonSerializerOptions`.

justinvp:
Make sense, thanks! There's already precedent for exposing System.Text.Json.JsonSerializerOptions from an API in our SDK: https://github.com/pulumi/pulumi-dotnet/blob/b03530027118e9305c2b69be37f88a70ceb81900/sdk/Pulumi/Core/Output.cs#L318

For now, another alternative workaround would be to get the config as a string value and the deserialize the JSON yourself using whatever deserializer/options you'd like.

rosieks:
Easy to implement that is just add new constructor that allows you to provide options. If that's acceptable way I can provide PR.

justinvp:
I don't think it'd be a new constructor. I think it'd be new overloads on the `Get*Object`/`Require*Object` methods:

```c#
public T GetObject<T>(string key); // Existing
public T GetObject<T>(string key, System.Text.Json.JsonSerializerOptions options); // New

public Output<T>? GetSecretObject<T>(string key); // Existing
public Output<T>? GetSecretObject<T>(string key, System.Text.Json.JsonSerializerOptions options); // New

public T RequireObject<T>(string key); // Existing
public T RequireObject<T>(string key, System.Text.Json.JsonSerializerOptions options); // New

public Output<T> RequireSecretObject<T>(string key); // Existing
public Output<T> RequireSecretObject<T>(string key, System.Text.Json.JsonSerializerOptions options); // New
```

> I can provide PR

That'd be great, thanks! But let me discuss the proposal with the team first in our design meeting early next week.

justinvp:
@rosieks, if you're still interested, we discussed the APIs I outlined above and are good add those, if you'd like to provide a PR. Please include tests. Thank you!

PR #373, opened 2024-11-08
Extending Config class to support passing JsonSerializerOptions

rosieks:
This PR address #370 

Extends `Config` class by providing `JsonSerializerOptions` overload to methods: GetObject, RequireObject, GetSecretObject, RequireSecretObject

### DIFF
--- a/sdk/Pulumi/Config.cs
+++ b/sdk/Pulumi/Config.cs
@@ -174,7 +174,7 @@ namespace Pulumi
         /// it to <see cref="JsonSerializer.Deserialize{TValue}(string, JsonSerializerOptions)"/>.
         /// </summary>
         [return: MaybeNull]
-        public T GetObject<T>(string key, JsonSerializerOptions? options)
+        public T GetObject<T>(string key, JsonSerializerOptions options)
             => GetObjectImpl<T>(key, nameof(GetSecretObject), options: options);
 
         /// <summary>
@@ -193,11 +193,11 @@ namespace Pulumi
         /// </summary>
         public Output<T>? GetSecretObject<T>(string key, JsonSerializerOptions options)
         {
-            var v = GetImpl(key);
+            var v = GetObjectImpl<T>(key, options: options);
             if (v == null)
                 return null;
 
-            return Output.CreateSecret(GetObjectImpl<T>(key, options: options)!);
+            return Output.CreateSecret(v);
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


sdk/Pulumi/Config.cs
```diff
@@ -177,6 +186,20 @@ private T GetObjectImpl<T>(string key, string? use = null, [CallerMemberName] st
...
+        /// or null if it doesn't exist. This works by taking the value associated with <paramref
+        /// name="key"/> and passing it to <see cref="JsonSerializer.Deserialize{TValue}(string, JsonSerializerOptions)"/>.
+        /// </summary>
+        public Output<T>? GetSecretObject<T>(string key, JsonSerializerOptions options)
+        {
+            var v = GetImpl(key);
+            if (v == null)
+                return null;
+
+            return Output.CreateSecret(GetObjectImpl<T>(key, options: options)!);
```
jkerken, line 200: This retrieves the value twice. Once from GetImpl and once with GetObjectImpl. 

```
var v = GetObjectImpl<T>(key, options: options);
if (v == null)
  return null;
return Output.CreateSecret(v);
```

This is missing a `var v = GetObjectImpl<T>(key, options: options);
if (v == null)
  return null;
 return Output.CreateSecret(v);`

sdk/Pulumi/Config.cs
```diff
@@ -155,19 +155,28 @@ private static Output<T> MakeStructSecret<T>(T value) where T : struct
...
             }
         }
 
+        /// <summary>
+        /// Loads an optional configuration value, as an object, by its key, or null if it doesn't
+        /// exist. This works by taking the value associated with <paramref name="key"/> and passing
+        /// it to <see cref="JsonSerializer.Deserialize{TValue}(string, JsonSerializerOptions)"/>.
+        /// </summary>
+        [return: MaybeNull]
+        public T GetObject<T>(string key, JsonSerializerOptions? options)
```
jkerken, line 177: GetObeject without the JsonSerializerOptions parameter is already present so it would be better to use a non null able argument here similiar how you did GetSecretObject

sdk/Pulumi/Config.cs
rosieks: Fixed

sdk/Pulumi/Config.cs
rosieks: Fixed

RonnyAnc:
Hi there,

Thank you both for working and proposing a solution for this, is anything missing? Can I help with something? It would be great to have this available.

Thank you so much!

jkodroff:
@pulumi-eon Why aren't the tests running here?

Frassle (approved):

PR #375, opened 2024-11-11
Always send a computed property value for unknown values

jkerken:
For some reason there was a difference between how an Unkown Output with Deps and without was serialized.

This causes  Unkown Outpust with Deps to be desearilized into null values if read back.

This pr fixes this.

### DIFF
--- a/.changes/unreleased/bug-fixes-375.yaml
+++ b/.changes/unreleased/bug-fixes-375.yaml
@@ -1,0 +1,6 @@
+component: sdk
+kind: bug-fixes
+body: Always send a computed property value for unknown values
+time: 2024-11-11T11:59:17.1828021+01:00
+custom:
+    PR: "375"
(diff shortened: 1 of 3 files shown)


Frassle:
So `OutputReference`'s `Value` property is marked nullable as a way to indicate unknowness rather than using `Computed`, so I don't think at first glance this is actually a bug? In fact if anything we should probably have a normalisation step in the `OutputReference` constructor so that if you try to create it with `Computed` it normalises back to just null.

jkerken:
@Frassle I can also change it like that, if that is the preferred way.

The problem is then these lines in the property value serializer deserialized method:
https://github.com/pulumi/pulumi-dotnet/blob/b03530027118e9305c2b69be37f88a70ceb81900/sdk/Pulumi/Provider/PropertyValueSerializer.cs#L601

Its checking there that the value is != null 

and then the output reference end up in :
` var deserialized = DeserializeValue(value, elementType, path);
  var outputDataValue = createOutputData.Invoke(new object?[]
  {
      ImmutableHashSet<Resource>.Empty,
      deserialized,
      true,
      false
  });
  `
  Where its turned into an Kown Null Input

Frassle:
Yeh can see this isn't quite lining up right on all paths, and found a few other odd bits missing while looking here. I think safe to leave it with me, I'll get a fix in and ping you on that PR so you can check it fixes the exact problem your hitting.

Frassle:
I've just hit https://github.com/pulumi/pulumi-dotnet/pull/381 to merge, see if that fixes your issue.

jkerken:
@Frassle Seems to fix the issue. Thanks closing this one

PR #384, opened 2024-11-13
Update xml docs on Release builds

Frassle:
We normally do Release builds as part of the workflows, and as part of `dotnet run build-sdk` so make sure the docs get updated for those builds.


Frassle:
Note that commit https://github.com/pulumi/pulumi-dotnet/pull/384/commits/d34131ec39cbf32839b84521a122f5d1f172c1a1 failed workflow checks because I hadn't updated the XML files in that commit. So we should now keep these up-to-date as merges will now fail if the docs aren't updated.

julienp (approved):

PR #385, opened 2024-11-13
DeferredOutput<T> 

Frassle:
Adds `DeferredOutput<T>` that allows the creation of an `Output<T>` that can later be resolved to the final result of a different `Output<T>`.

This can be used to solve partial circular reference problems. For example given two components that create multiple resources and return some results from that, you would normally have to have all the inputs for one component available to call it and then afterwards call the other component. With an `DeferredOutput<T>` you can instead set up a circular reference:
```csharp
var ocs = new DeferredOutput<int>()
var (resultA1, resultA2) = MakeComponentA(ocs.Output);
var resultB = MakeComponentB(resultA1);
ocs.Resolve(resultB);
```
As long as `MakeComponentA` can resolve its A1 result without needing the incoming output resolved then the above will work.

Note that it's possible to deadlock the program with this feature. If resultA1 above can't resolve until the input is resolved, and resultB can't resolve till its input is resolved then both will end up stuck waiting for the other. Users can already deadlock their programs with other dotnet constructs (for example doing a similar thing with `TaskCompletionSource<T>`).

### DIFF
--- a/.changes/unreleased/Improvements-385.yaml
+++ b/.changes/unreleased/Improvements-385.yaml
@@ -1,0 +1,6 @@
+component: sdk
+kind: Improvements
+body: Add `DeferredOutput` for resolving some output/input cycles
+time: 2024-11-19T08:28:36.110409Z
+custom:
+  PR: "385"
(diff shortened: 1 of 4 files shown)


Zaid-Ajaj (commented):
Looks good to start using in C# program-gen, unless there are things we are uncertain about yet?

Frassle:
Yeh let me fill in docs then we can probably merge this.

lunaris (approved):

Zaid-Ajaj:
> Are we settled on the API design and naming for this? What's it going to look like in the other SDKs?

Will be discussing next week but leaning towards a unified name `DeferredOutput<T>` for all the languages

Zaid-Ajaj (approved):

PR #386, opened 2024-11-13
Run tests with .NET 9.0

julienp:
Add .NET 9.0 to the CI test matrix


julienp:
@Frassle found the issue https://github.com/dotnet/sdk/issues/44610

julienp:
Workaround in CLI https://github.com/pulumi/pulumi/pull/17783

Once that's released, we can do the upgrade here.

justinvp:
We can close this now. .NET 9 was added to the matrix in https://github.com/pulumi/pulumi-dotnet/pull/500.

PR #390, opened 2024-11-15
Fix the language plugin to return a version

### LINKED ISSUE #389: `pulumi about` always shows `unknown` for the dotnet language plugin version
When running `pulumi about` for a .NET Pulumi program, it shows `unknown` for the language plugin version. Instead, it should show the version of the language plugin.

`GetPluginInfo` is returning `version.Version` for the version:

https://github.com/pulumi/pulumi-dotnet/blob/e105d3278fea40be1f3008e36642756b0085e981/pulumi-language-dotnet/main.go#L805-L809

But the wrong package is being used. It's using `"github.com/pulumi/pulumi/sdk/v3/go/common/version"` when it should be using `"github.com/pulumi/pulumi-dotnet/pulumi-language-dotnet/version"` as the latter package's `Version` variable is the one that's actually being set during build:

https://github.com/pulumi/pulumi-dotnet/blob/e105d3278fea40be1f3008e36642756b0085e981/.goreleaser.yml#L17-L18

justinvp:
This change fixes `pulumi about` to show the real version instead of `unknown` for the dotnet language plugin.

The fix updates [`GetPluginInfo`](https://github.com/pulumi/pulumi-dotnet/blob/e105d3278fea40be1f3008e36642756b0085e981/pulumi-language-dotnet/main.go#L805-L809) to return the `Version` from the right `version` Go package&mdash;the one that's actually being set during build via `ldflags` (`"github.com/pulumi/pulumi-dotnet/pulumi-language-dotnet/version"` instead of `"github.com/pulumi/pulumi/sdk/v3/go/common/version"`, which looks to be a hold over from when this code lived in the pulumi/pulumi repo).

Fixes #389


tgummerer (approved):

justinvp:
I added some commits to set a version when building locally. Going to merge, but happy to address any feedback subsequently.

PR #396, opened 2024-11-20
Fix RunPlugin and CI jobs

Frassle:
Since https://github.com/pulumi/pulumi/pull/17763 the pwd property is the working directory, not the program directory. We need to fix up RunPlugin to use both directories correctly, loading the csproj from the program directory, but then setting the working directory to pwd.

This also updates the CI jobs to always use the latest version of the CLI. We were running into issues because different runners happened to have different CLI versions installed.

For now I've also disabled fail-fast on the integration tests to make re-running them with flaky tests more likely to pass.


Frassle:
To see if we can work out what CI jobs are failing

julienp (approved):
LGTM

PR #404, opened 2024-11-21
Add Parameterize to the provider interface

Frassle:
This adds the new Parameterize method to the dotnet provider SDK.

This is tested via a little integration test. I've had to update the C# testprovider to handle being parameterized, and have written a small (incomplete) schema by hand to test gen-sdk works with this. This required renaming the echo resources property because C# can't handle the resource and property both being called echo. That required updating some other tests slightly to accommodate the rename.


sdk/Pulumi/Provider/Provider.cs
```diff
@@ -19,9 +19,81 @@
 using Microsoft.AspNetCore.Hosting.Server;
 using Microsoft.AspNetCore.Hosting.Server.Features;
 using Microsoft.AspNetCore.Server.Kestrel.Core;
+using System.Reflection.Metadata;
 
 namespace Pulumi.Experimental.Provider
 {
+    public abstract class Parameters { }
```
Frassle, line 26: I wasn't sure if it was best to go with this abstract base class and two concrete classes for the different parameter types, or to just have two nullable fields on ParameterizeRequest for each choice. This feels more correct, because `parameters` has to be one or the other, while the two nullable fields _technically_ could both be filled in. But also a bit less obvious to use compared to just two fields. 
For what it's worth Go also does something like this abstract class trick for the parameters.

Zaid-Ajaj (approved):
Neat 🎉

integration_tests/parameterized/Parameterized.csproj
```diff
@@ -0,0 +1,14 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <ItemGroup>
+    <ProjectReference Include="sdk\dotnet\Pulumi.Pkg.csproj" />
+  </ItemGroup>
+
+  <PropertyGroup>
+    <DefaultItemExcludes>$(DefaultItemExcludes);sdk/**/*.cs</DefaultItemExcludes>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net6.0</TargetFramework>
```
justinvp, line 10: This will need to be updated to match updates to other integration test projects in https://github.com/pulumi/pulumi-dotnet/pull/410

The macos-13 runner doesn't come with .NET 6 installed. So when running the .NET 8 matrix, the integration tests were failing because they target .NET 6. This change allows it to work with only .NET 8 installed, because the GHA workflow sets `TARGET_FRAMEWORK` envvar based on the matrix .NET version.

```suggestion
    <TargetFramework Condition=" '$(TARGET_FRAMEWORK)' != '' ">$(TARGET_FRAMEWORK)</TargetFramework>
    <TargetFramework Condition=" '$(TARGET_FRAMEWORK)' == '' ">net6.0</TargetFramework>
```

integration_tests/testprovider/TestProvider.csproj
```diff
@@ -2,7 +2,8 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework>net8.0</TargetFramework>
```
justinvp, line 5: Similar to my other comment above.

Note: I was planning to update the fallback value for when the envvar isn't set from `net6.0` to `net8.0` in a follow-up change.

```suggestion
    <TargetFramework Condition=" '$(TARGET_FRAMEWORK)' != '' ">$(TARGET_FRAMEWORK)</TargetFramework>
    <TargetFramework Condition=" '$(TARGET_FRAMEWORK)' == '' ">net6.0</TargetFramework>
```

PR #407, opened 2024-11-22
await background tasks in inline deployments

jkerken:


sdk/Pulumi/Deployment/Deployment_Inline.cs
```diff
@@ -80,7 +80,7 @@ internal static async Task<T> RunInlineAsyncWithResult<T>(IDeploymentBuilder bui
                 {
                     try
                     {
-                        var result = await runnerFunc(runner).ConfigureAwait(false);
+                        var result = await runner.RunAsync(()=> runnerFunc(runner)).ConfigureAwait(false);
```
Frassle, line 83: I don't think we actually need to pass the runner to runnerFunc here. It's not used in the provider implementation.

PR #417, opened 2024-12-03
Generalize and fix some workflow code

Frassle:
Fix some 'if' checks to be more general, and some job descriptions that incorrectly said they were installing .NET 6.


julienp (approved):
LGTM

PR #429, opened 2024-12-12
Remove TestBuildDll that is not running

Zaid-Ajaj:
While working on .NET conformance tests, encountered this test that is not running in CI and is not working because of a missing `testdata/build-dll` directory. Removing it for now and adding a working version later cc @tgummerer


tgummerer:
Huh, why was this not running in CI? :thinking: 

I still have the build-dll directory around locally, so I can submit a fix for the test

Frassle (approved):

Zaid-Ajaj:
Replaced by #430

PR #314, opened 2024-08-07
Implement .NET conformance testing

Zaid-Ajaj:
This PR is an initial implementation of .NET conformance tests. It extends pulumi-language-dotnet with `GenerateProject`, `GeneratePackage` and `Pack` 

It references a specific Pulumi [commit](https://github.com/pulumi/pulumi/commit/9fd37d48612be032ccb054e9e856ef04cc9cd9b7) which includes  https://github.com/pulumi/pulumi/pull/18026 that is not released yet. 

Also includes https://github.com/pulumi/pulumi/pull/18042 from [this commit](https://github.com/pulumi/pulumi/commit/109d75b57c54db972a3a02b8bead4220ae870059)

PR #430, opened 2024-12-12
add missing build-dll directory

tgummerer:
This should fix the TestBuildDll test.

We should still make sure this test is acutually running in CI.

Replaces https://github.com/pulumi/pulumi-dotnet/pull/429

/cc @Zaid-Ajaj

### DIFF
--- a/pulumi-language-dotnet/testdata/build-dll/.gitignore
+++ b/pulumi-language-dotnet/testdata/build-dll/.gitignore
@@ -1,0 +1,5 @@
+/.pulumi/
+[Bb]in/
+[Oo]bj/
+
+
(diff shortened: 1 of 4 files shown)


julienp:
We only run seem to run integration tests in CI https://github.com/pulumi/pulumi-dotnet/blob/b93b3c23e619e660453f6d26b8d6a89144ba0523/.github/workflows/pr.yml#L72, and not the tests in `./pulumi-language-dotnet` 🫤

Zaid-Ajaj:
> We should still make sure this test is acutually running in CI.

Doing this already in https://github.com/pulumi/pulumi-dotnet/pull/314 so it should be fine

Zaid-Ajaj (approved):

PR #434, opened 2024-12-17
Add support for unions to PropertyValueSerializer

jkerken:
Using `InputUnion` or `Union` in Resource args for an MLC did not work because of missing support in the PropertyValueSerializer 

Added in this pr


Frassle:
Also check the code formatting. `dotnet run format-sdk`

PR #437, opened 2024-12-20
Tolerate plugins with warnings

Frassle:
Found this while on my adventures with dotnet plugins. If your plugin project has any warnings it's unusable in shimless mode (i.e. `RunPlugin`) because msbuild would first print the warning then start the plugin and print the port number. This lead to errors like:
```
error:  plugin [E:\VisualStudio\Projects\pulumi\jsonschema\Provider\pulumi-resource-Provider] wrote an invalid port to stdout: could not parse port: strconv.Atoi: parsing "E:\\VisualStudio\\Projects\\pulumi\\jsonschema\\Provider\\Provider.fsproj : warning NU1903: Package 'System.Text.Json' 6.0.2 has a known high severity vulnerability, https://github.com/advisories/GHSA-8g4q-xg66-9fp4": invalid syntax
```

When running shimless these warnings don't really matter. So this PR rewrites `RunPlugin` to do the build separately, and then invoke `dotnet run` with the `--no-build` flag. I would have preferred to _just_ use `dotnet run` but there doesn't seem to be anyway to pass it flags to silence the msbuild output from it. 

This is tested by adding a warning to the testprovider project that always triggers. As we use the testprovider project in shimless mode in the integration tests this shows that a project with warnings can still be used.

### DIFF
--- a/.changes/unreleased/Improvements-437.yaml
+++ b/.changes/unreleased/Improvements-437.yaml
@@ -1,0 +1,6 @@
+component: runtime
+kind: Improvements
+body: Plugins with msbuild warnings can still be run
+time: 2024-12-20T22:28:08.41871359Z
+custom:
+  PR: "437"
--- a/integration_tests/integration_dotnet_test.go
+++ b/integration_tests/integration_dotnet_test.go
@@ -32,7 +32,6 @@ import (
 	"github.com/pulumi/pulumi/pkg/v3/testing/integration"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/apitype"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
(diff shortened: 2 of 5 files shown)


Frassle:
Seems something is off with the tests not using the changes in the language host, works fine locally.

julienp:
> Seems something is off with the tests not using the changes in the language host, works fine locally.

Looking at CI, I don't think we're building the language runtime in the integration tests 😬  I thought I noticed that earlier and fixed it, but maybe that was for java?

julienp:
Ah no, we build in the makefile for `test_integration`, but I guess the built binary is not in the path?

Frassle:
> but I guess the built binary is not in the path?

No it's not. Most of the integration tests change the PATH env to pick up the right thing but in the places we do `NewEnvironment` we're not getting that edit.

Issue #438, opened 2024-12-30
Make OutputConstructorAttribute optional


rosieks:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Make `OutputConstructorAttribute` mandatory only if there is more than one constructor. The goal is to make make it easy to use records with primary constructor like that:

```csharp
[OutputType]
public record SomeOutput(string Value1, string Value2);
```

Right now such code throw exception like this:

```
System.InvalidOperationException: SomeOutput had [OutputTypeAttribute], but did not contain constructor marked with [OutputConstructorAttribute].
```

### Affected area/feature

SDK / Serialization

PR #439, opened 2024-12-30
Make OutputConstructorAttribute optional

rosieks:
This PR address issue #438 
It changes behavior of converter in a way that if there is no `OutputConstructorAttribute` and at the same time there is single public constructor then it tries to use it.


sdk/Pulumi/Serialization/Converter.cs
```diff
@@ -499,7 +499,19 @@ static bool CheckEnumType(Type targetType, Type underlyingType)
...
+            if (outputConstructor == null)
+            {
+                var publicConstructors = constructors.Where(x => x.IsPublic).ToArray();
+                if (publicConstructors.Length == 1)
+                {
+                    return publicConstructors[0];
+                }
+            }
+
+            return outputConstructor;
```
brandonpollack23, line 514: I personally prefer the early return style the other way with explicit null return:

```csharp
if (outputConstructor != null)
{
  return outputConstructor;
}

var publicConstructors = constructors.Where(x => x.IsPublic).ToArray();
if (publicConstructors.Length == 1)
{
  return publicConstructors[0];
}

return null;
```

brandonpollack23 (changes requested):

sdk/Pulumi/Serialization/Converter.cs
rosieks: Changed

brandonpollack23 (approved):

Frassle:
Hi @rosieks, if you rebase/merge main into this I think it will retrigger the CI checks and we can get this merged.

PR #440, opened 2025-01-02
[language-runtime] Implement GetRequiredPackages to replace GetRequiredPlugins

### LINKED ISSUE #432: Implement GetRequiredPackages in the language host
Implement `GetRequiredPackages` in the .NET language host, needed for https://github.com/pulumi/pulumi/issues/17507.

Zaid-Ajaj:
### Description

The .NET language runtime currently uses `GetRequiredPlugins` to determine which resource plugins a program needs in order to run. However, `GetRequiredPlugins` doesn't account for the parameterization of the plugins. This is where `GetRequiredPackages` comes into play which allows the language runtime to return required plugins/packages along with their parameterization info (loaded from `pulumi-plugin.json` files). 

This also allows the language runtime to correctly work with `pulumi import` for parameterized providers. 

Fixes #432 and unskips tests:
 - l2-parameterized-resource
 - l2-explicit-parameterized-provider

> Note: the tests were failing locally but that was because the local package nugets used in the conformance tests had old contents and needed to be invalidated. This so that `pulumi-plugin.json` could be filled with the parameterization values.


tgummerer (approved):
Where do these `logo.png` files here come from?  Do we really need them?

Zaid-Ajaj:
> Where do these logo.png files here come from? Do we really need them?

@tgummerer Technically not required but they are part of SDK gen. We include them to change the icon on nuget.org for Pulumi packages

tgummerer:
> We include them to change the icon on nuget.org for Pulumi packages

Thanks, TIL

PR #444, opened 2025-01-09
Wait for resources in the input property dependencies

julienp:
Besides the explicitly provided resources from the `dependsOn` option, we need to also take into account the resource dependencies from the invoke’s arguments.

https://github.com/pulumi/pulumi/issues/17747

### DIFF
--- a/sdk/Pulumi/Deployment/Deployment_Invoke.cs
+++ b/sdk/Pulumi/Deployment/Deployment_Invoke.cs
@@ -156,6 +156,11 @@ namespace Pulumi
 
             // Resource dependencies that will be added to the Output
             var resourceDependencies = new HashSet<Resource>();
+            // If we depend on any CustomResources, we need to ensure that their
+            // ID is known before proceeding. If it is not known, we will return
+            // an unknown result.
+            // Add the dependencies from the inputs to the set of resources to wait for.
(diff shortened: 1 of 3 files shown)


sdk/Pulumi/Deployment/Deployment_Invoke.cs
```diff
@@ -156,35 +156,41 @@ private async Task<OutputData<T>> RawInvoke<T>(
 
             // Resource dependencies that will be added to the Output
             var resourceDependencies = new HashSet<Resource>();
+            // If we depend on any CustomResources, we need to ensure that their
+            // ID is known before proceeding. If it is not known, we will return
+            // an unknown result.
+            // Add the dependencies from the inputs to the set of resources to wait for.
+            var resourcesToWaitFor = serializedArgs.PropertyToDependentResources.Values.SelectMany(r => r).ToHashSet<Resource>();
```
brandonpollack23, line 163: Is there a more clear function in C# like Flatten or some such?

brandonpollack23 (approved):
might wanna wait for someone else who is more familiar but makes sense to me as is I think

sdk/Pulumi/Deployment/Deployment_Invoke.cs
julienp: Yeah, I am not a C# expert either and I am also more used to this being called `flatten` or something similar. It seems that's the way to do it. When I search for "c# flatten list of lists", most results tell me to use `SelectMany`.

PR #445, opened 2025-01-16
Fix dotnet version

julienp:
With the update to Ubuntu 24.04 for the runners, the preinstalled .NET versions have changed. This revealed some issues with how we select the .NET version to use in the matrix tests. The `actions/setup-dotnet` action does install the version we want, but the dotnet command still use the runner’s default version.

We have to create a `global.json` to make it use the desired version.

https://github.com/actions/setup-dotnet?tab=readme-ov-file#matrix-testing

### DIFF
--- a/integration_tests/testprovider/TestProvider.cs
+++ b/integration_tests/testprovider/TestProvider.cs
@@ -46,40 +46,40 @@ public class TestProvider : Provider
 
     public override Task<GetSchemaResponse> GetSchema(GetSchemaRequest request, CancellationToken ct)
     {
-        var schema = """
+        var schema = @"
 {
-    "name": "NAME",
-    "version": "1.0.0",
-    "resources": {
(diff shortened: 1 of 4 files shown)


integration_tests/testprovider/TestProvider.cs
```diff
@@ -46,40 +46,40 @@ public override Task<ParameterizeResponse> Parameterize(ParameterizeRequest requ
 
     public override Task<GetSchemaResponse> GetSchema(GetSchemaRequest request, CancellationToken ct)
     {
-        var schema = """
```
julienp, line 49: Raw strings `"""..."""` were introduced in C# 11, however .NET 6.0 only supports up to C# 10. Convert this to use verbatim strings `@"..."`, which unfortunately require double quoting quotes.

lunaris (approved):

PR #453, opened 2025-02-05
Component schema and runtime inference prototype

mikhailshilkov:


integration_tests/provider_construct/testcomponent-dotnet/Component.cs
```diff
@@ -50,10 +50,13 @@ class Component : ComponentResource
     public Output<ComplexType> ComplexResult { get; set; }
 
     public Component(string name, ComponentArgs args, ComponentResourceOptions? opts = null)
-        : base("test:index:Test", name, args, opts)
```
Frassle, line 53: Please duplicate this folder, it's useful having a low level test for components as well.

PR #459, opened 2025-02-12
Handle null runtime values in InputMap/List cast

Frassle:
Fixes https://github.com/pulumi/pulumi-dotnet/issues/456

This is a bit of a sad fix but `InputMap` used to tolerate the implicit cast from dictionary being called with null by just passing that null down. If you tried to call Add or enumerate it would NRE, but if it just got assigned to another resource property the serialiser could handle it.

https://github.com/pulumi/pulumi-dotnet/pull/449 broke this because it always tried to enumerate the incoming immutable dictionary to build the new internal nested immutable dictionary. This fixes it to behave as before were null flows through as null.

### DIFF
--- a/.changes/unreleased/bug-fixes-459.yaml
+++ b/.changes/unreleased/bug-fixes-459.yaml
@@ -1,0 +1,6 @@
+component: sdk
+kind: bug-fixes
+body: Handle null in InputMap/List implicit conversions
+time: 2025-02-12T16:47:02.408546Z
+custom:
+    PR: "459"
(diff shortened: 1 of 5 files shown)


justinvp:
Does `InputList` have a similar issue?

Frassle:
Maybe, ImmutableArray can't be null but its default value is invalid I think. I'll check.

Frassle:
Added some more tests and fixed up InputList as well.

lunaris (approved):

Issue #455, opened 2025-02-11
ResourceTransforms is not working at all


Meydan-Ozeri:
### What happened?

I have an issue with ResourceTransforms option in the ComponentResourceOptions of a Pulumi resource.
No matter what i put inside it never runs, no log is even appearing in the console and no change i try to make is reflected.

The regular ResourceTransformations (which as i read will be deprecated in the future) does run successfully, but it doesn't iterate over any child resources of the Helm Charts, only the top level "chart" resource is appearing there.

### Example

I have this defined on a Helm Chart V4 resource:
``
ResourceTransforms =
{
	async (args, _) =>
	{
		Console.WriteLine("Transforming resource");
		if (args.Type.Contains("Deployment"))
		{
			return new ResourceTransformResult(
				args.Args,
				CustomResourceOptions.Merge((CustomResourceOptions)args.Options, new() { IgnoreChanges = ["spec.replicas"] })
			);
		}
		return null;
	}
}
``

### Output of `pulumi about`

running 'dotnet build -nologo'
  Determining projects to restore...

  All projects are up-to-date for restore.

  Deployment -> C:\Meydan\Projects\TradingPro\Deployment\bin\Debug\net9.0\Deployment.dll

Build succeeded.

    0 Warning(s)
    0 Error(s)

Time Elapsed 00:00:01.47

'dotnet build -nologo' completed successfully
Enter your passphrase to unlock config/secrets
    (set PULUMI_CONFIG_PASSPHRASE or PULUMI_CONFIG_PASSPHRASE_FILE to remember):
Enter your passphrase to unlock config/secrets
CLI
Version      3.149.0
Go Version   go1.23.6
Go Compiler  gc

Plugins
KIND      NAME                     VERSION
resource  command                  1.0.1
resource  docker                   4.6.1
resource  docker-build             0.0.10
language  dotnet                   3.73.0
resource  gcp                      8.18.0
resource  kind                     0.0.5
resource  kubernetes               4.21.1
resource  kubernetes-cert-manager  0.1.1

Host
OS       Microsoft Windows 11 Pro
Version  10.0.26100 Build 26100
Arch     x86_64

This project is written in dotnet: executable='C:\Program Files\dotnet\dotnet.exe' version='9.0.102'

Current Stack: organization/SetupKubernetesCluster/Development

TYPE                                                           URN
...

Found no pending operations associated with Development

Backend
Name           Meydan
URL            file://./
User           MEYDAN\meyda
[...]

PR #460, opened 2025-02-12
Fix Provider and Providers in transforms

Frassle:
Fixes https://github.com/pulumi/pulumi-dotnet/issues/455

When marshalling options back to the engine after resource transforms we were incorrectly sending `provider` and `providers` back as just the URNs. They should in fact be "refs" of the form `<urn>::<id>`.


lunaris (approved):
Does this need a test or is it covered?

Frassle:
> Does this need a test or is it covered?

It's not but I didn't work out a quick way to test it the other day when writing this. I want to have another look, but it's probably worth merging this now to get the fix in.

PR #461, opened 2025-02-12
Add ability to disable ToString on Output<T>

JasonWhall:
This PR provides an ability that is already present in the Python and JS languages to to disable converting `Output<T>` to a string. 

Currently in the dotnet language this will return a helpful message in the string contents to advise not to do this but does not fail the pulumi program.

This functionality is opted-in by setting the environment variable `PULUMI_ERROR_OUTPUT_STRING` to `true` or `1` to throw an exception.


justinvp (approved):
LGTM

Thanks, @JasonWhall!

Issue #458, opened 2025-02-12
InputMap doesn't allow the same value to be added anymore


Frassle:
### What happened?

InputMap used to allow the same key and value to be added multiple times. Now it errors because internally the value is wrapped in an Input now and so doesn't have a simple equality. 

### Example

See https://github.com/pulumi/pulumi-dotnet/pull/449#issuecomment-2653414395

### Output of `pulumi about`

N/A

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

PR #462, opened 2025-02-13
Fix adding duplicate key-value pairs to InputMap

Frassle:
Fixes https://github.com/pulumi/pulumi-dotnet/issues/458

Before https://github.com/pulumi/pulumi-dotnet/pull/449 it used to be possible to `Add` the same key value pair to `InputMap` multiple times. This because of a feature of the underlying ConcurrentDictionary which will use `EqualityComparer` to see if the key and value being added are the same as what's in the dictionary already and if so just returns the existing dictionary. 

#449 broke this because now all the values of the dictionary are `Input`s which don't have a sensible notion of equality. This adds some extra handling in our Add method to try and maintain this feature.

It _does_ have two oddities:
1. The exception is recorded on the value, not the map overall. This shouldn't be observable because users can only resolve InputMap as a single `Input<Dict<K, V>>` and not directly access the nested input values. But it's worth calling out to keep in mind for the future.
2. The value always merges the secretes and dependencies of all `Input`s added to the one key. Those properties don't get overwritten, so if you add `("k", secret("hello"))` then add `("k", "hello")` you might expect that value to no longer be secret because it's been overwritten, but in fact the secretness is merged and so it's still secret. Fixing this would require a custom apply implementation to await both values, then _iff_ they're both known keep just the output data from the later one, but you'd still end up with merged data if either was unknown. All in all it didn't feel worth the complexity to do different here.

### DIFF
--- a/sdk/Pulumi.Tests/Core/InputTests.cs
+++ b/sdk/Pulumi.Tests/Core/InputTests.cs
@@ -1,5 +1,6 @@
 // Copyright 2016-2019, Pulumi Corporation
 
+using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Threading.Tasks;
@@ -141,6 +142,31 @@ namespace Pulumi.Tests.Core
                 sample = new SampleArgs { Union = Union<string, int>.FromT1(456) };
                 data = await sample.Union.ToOutput().DataTask.ConfigureAwait(false);
                 Assert.Equal(456, data.Value);
+            });
+
+        [Fact]
+        public Task InputMapAdd()
+            => RunInPreview(async () =>
+            {
+                var map = new InputMap<string>();
+
+                // Add a key and value
+                map.Add("K1", "V1");
+                // Add the a new key and the same value twice
+                map.Add("K2", "V2");
+                map.Add("K2", "V2");
+
+                // We should be able to get this map still
+                var data = await map.ToOutput().DataTask.ConfigureAwait(false);
+                Assert.Equal(2, data.Value.Count);
+                Assert.Equal(new Dictionary<string, string> { ["K1"] = "V1", ["K2"] = "V2" }, data.Value);
+
+                // Add a new key and two different values
+                map.Add("K3", "V3");
+                map.Add("K3", "V3_wrong");
+
+                // This should now throw an exception
+                await Assert.ThrowsAsync<ArgumentException>(() => map.ToOutput().DataTask).ConfigureAwait(false);
             });
 
         private class SampleArgs
--- a/sdk/Pulumi/Core/InputMap.cs
+++ b/sdk/Pulumi/Core/InputMap.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
 using System.Threading;
+using Microsoft.VisualBasic;
 
 namespace Pulumi
 {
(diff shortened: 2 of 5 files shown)


sdk/Pulumi/Core/InputMap.cs
```diff
@@ -91,7 +92,30 @@ private InputMap(Input<ImmutableDictionary<string, Input<V>>> values)
 
         public void Add(string key, Input<V> value)
         {
-            Value = Value.Apply(self => self.Add(key, value));
+            Value = Value.Apply(self =>
+            {
+                // ImmutableDictionary allows the same value to be added twice for the same. Sameness is decided via EqualityComparer<V>,
```
brandonpollack23, line 97:
```suggestion
                // ImmutableDictionary allows the same value to be added twice for the same key. Sameness is decided via EqualityComparer<V>,
```

sdk/Pulumi/Core/InputMap.cs
```diff
@@ -91,7 +92,30 @@ private InputMap(Input<ImmutableDictionary<string, Input<V>>> values)
...
+
+                // Else we can only know if the key is ok to add if we compare the values inside the Input<T>. This is a
+                // bit odd, because firstly the exception is only seen if awaiting _this_ value, not the value for the
+                // dictionary itself. That shouldn't be an issue because the only thing that can resolve the
+                // dictionaries inner levels separately is the property serializer and it always reads all the values.
+                // The other oddity is this merges the secretness and dependency information of the two values, so if
+                // you add "K" with a secret "hello", and then call add for "K" again with a non-secret "hello" it's going to
+                // merge both and keep the secret tag. Doing better then this requires a custom Apply function here.
+                return self.SetItem(key,
+                    Output.Tuple(existingValue, value).Apply(x =>
```
brandonpollack23, line 115: nit: can we destructure the [tuple](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/deconstruct)

sdk/Pulumi/Core/InputMap.cs
```diff
@@ -91,7 +92,30 @@ private InputMap(Input<ImmutableDictionary<string, Input<V>>> values)
...
+                    return self.Add(key, value);
+                }
+
+                // Else we can only know if the key is ok to add if we compare the values inside the Input<T>. This is a
+                // bit odd, because firstly the exception is only seen if awaiting _this_ value, not the value for the
+                // dictionary itself. That shouldn't be an issue because the only thing that can resolve the
+                // dictionaries inner levels separately is the property serializer and it always reads all the values.
+                // The other oddity is this merges the secretness and dependency information of the two values, so if
+                // you add "K" with a secret "hello", and then call add for "K" again with a non-secret "hello" it's going to
+                // merge both and keep the secret tag. Doing better then this requires a custom Apply function here.
```
brandonpollack23, line 113: This comment is super long and hard for me to follow a bit, can we break it down into bullet points?

brandonpollack23 (approved):

sdk/Pulumi.Tests/Core/InputTests.cs
```diff
@@ -143,6 +144,31 @@ public Task InputUnionInitializer()
                 Assert.Equal(456, data.Value);
             });
 
+        [Fact]
```
brandonpollack23, line 147: It'd probably be a good idea to do a secret in the test as well, even if we have the overlap issue

sdk/Pulumi/Core/InputMap.cs
Frassle: Yes but not in the parameter list: https://github.com/dotnet/csharplang/discussions/258
So then this needs expanding to be a block lambda, and I'm not sure that's worth it given the context is all on the same line.

sdk/Pulumi/Core/InputMap.cs
Frassle: ack

sdk/Pulumi/Core/InputMap.cs
```diff
@@ -6,6 +6,7 @@
 using System.Collections.Immutable;
 using System.Linq;
 using System.Threading;
+using Microsoft.VisualBasic;
```
justinvp, line 9: huh

PR #465, opened 2025-02-14
Go mod tidy and change the gomod to use the local pulumi repository.

brandonpollack23:


brandonpollack23:
This was dumb, not everyone's path is set up like mine

Frassle:
Can copy the same description from https://github.com/pulumi/pulumi-java/pull/1671

We'll still need to update the submodule as the ref is always resolves as a commit not a branch? But it probably makes sense to build against master than a release tag.

brandonpollack23:
> Can copy the same description from [pulumi/pulumi-java#1671](https://github.com/pulumi/pulumi-java/pull/1671)
> 
> We'll still need to update the submodule as the ref is always resolves as a commit not a branch? But it probably makes sense to build against master than a release tag.

yes we'll have to manually bump it afaict, we may be able to have a routine which does it, but its probably better to just do it when we update this repo.  When we get around to adding pu/pu conformance testing for this repo itll be a forcing function

brandonpollack23:
we discussed this in ideation and decided to go with rennovate bot to update pulumi/pulumi submodule's pointer and the go.mod version to always point to the current master externally, as well as open a PR for it so all the tests (and consequently any new conformance tests) run and fail, letting us know to implement or bypass the test.

PR #466, opened 2025-02-14
Add conformance for rootDirectory pcl function

brandonpollack23:
Add conformance for projectRoot pcl function

Part of pulumi/pulumi#18565
Blocked by pulumi/pulumi#18595

Changes required to get project root in dotnet and have conformance
tests pass.


brandonpollack23:
good to go

brandonpollack23:
blocking PR is approved, just running tests so this should be good to review

sdk/Pulumi/Deployment/Deployment.cs
```diff
@@ -41,30 +41,39 @@ namespace Pulumi
...
+            new AsyncLocal<DeploymentInstance?>();
 
         /// <summary>
         /// The current running deployment instance. This is only available from inside the function
         /// passed to <see cref="Deployment.RunAsync(Action)"/> (or its overloads).
         /// </summary>
         public static DeploymentInstance Instance
         {
-            get => _instance.Value ?? throw new InvalidOperationException("Trying to acquire Deployment.Instance before 'Run' was called.");
+            get =>
```
Frassle, line 53: Don't reformat whitespace everywhere, just makes the diff noisy.

brandonpollack23: that shoudl be better, wasn't aware I had a strict formatter turned on

sdk/Pulumi/Deployment/Deployment.cs
```diff
@@ -119,6 +121,7 @@ private Deployment(RunnerOptions? runnerOptions = null)
             if (string.IsNullOrEmpty(monitor) ||
                 string.IsNullOrEmpty(engine) ||
                 string.IsNullOrEmpty(project) ||
+                string.IsNullOrEmpty(rootDirectory) ||
```
Frassle, line 124: You can't add this here. If someone uses a new SDK with the old language host it will error out on this.

sdk/Pulumi/Deployment/Deployment_Inline.cs
```diff
@@ -17,6 +17,7 @@ private Deployment(IDeploymentBuilder builder, InlineDeploymentSettings settings
 
             _organizationName = settings.Organization ?? "organization";
             _projectName = settings.Project;
+            _rootDirectory = "rootDirectory";
```
Frassle, line 20: We should probably make root directory nulllable. Both running against the old engine, and running inline programs won't have a root directory.

sdk/Pulumi/Deployment/Deployment.cs
brandonpollack23: okidoki

sdk/Pulumi/Deployment/Deployment_Inline.cs
brandonpollack23: done

sdk/Pulumi/Deployment/Deployment.cs
```diff
@@ -160,6 +163,7 @@ internal Deployment(IEngine engine, IMonitor monitor, TestOptions? options)
             _isDryRun = options?.IsPreview ?? true;
             _stackName = options?.StackName ?? "stack";
             _projectName = options?.ProjectName ?? "project";
+            _rootDirectory = options?.RootDirectory ?? "rootDirectory";
```
Frassle, line 166: Not sure this should default to anything? Like null is easier to match against than "rootDirectory'

brandonpollack23: I suppose so, the others aren't nullable but it kinda makes me think they should be as well

Frassle: Well the others always get set except org and "organisation" was I suppose a half sensible default choice for that, although in hindsight maybe that should have been nullable or throwing as well for old engines.

I think this probably _could_ be set for inline programs as well (the autoapi does make a temp folder for them to run in). So it's just what to do for old engines, which I think has three choices:
1) Be nullable, return null for old engines
2) Don't be nullable and throw a runtime exception if running against an old engine. We do this for other things like transforms
3) Don't be nullable and have a default value, but I'm not sure anything makes sense for that.

So actually thinking about it (and sorry to go back and forth on you like this) but maybe we should do option 2. Make the user API non-nullable, but internally track if it's actually set by the engine or not and if the user tries to access it throw an exception like we do if they try and register a stack transform and it isn't supported: `throw new Exception("The Pulumi CLI does not support root directory. Please update the Pulumi CLI.");`

brandonpollack23: Yup, agreed

brandonpollack23: Ok this is done now

sdk/Pulumi/Deployment/Deployment.cs
```diff
@@ -120,6 +122,7 @@ private Deployment(RunnerOptions? runnerOptions = null)
                 string.IsNullOrEmpty(engine) ||
                 string.IsNullOrEmpty(project) ||
                 string.IsNullOrEmpty(stack) ||
+                string.IsNullOrEmpty(rootDirectory) ||
```
Frassle, line 125:
```suggestion
```

This can't be here it will cause new SDKs to error against old engines

sdk/Pulumi/Deployment/Deployment.cs
```diff
@@ -160,6 +164,7 @@ internal Deployment(IEngine engine, IMonitor monitor, TestOptions? options)
             _isDryRun = options?.IsPreview ?? true;
             _stackName = options?.StackName ?? "stack";
             _projectName = options?.ProjectName ?? "project";
+            _rootDirectory = options?.RootDirectory ?? "rootDirectory";
```
Frassle, line 167:
```suggestion
            _rootDirectory = options?.RootDirectory;
```

sdk/Pulumi/Deployment/Deployment_Inline.cs
Frassle, line 20:
```suggestion
            _rootDirectory = settings.RootDirectory;
```

sdk/Pulumi/Deployment/IDeployment.cs
```diff
@@ -16,6 +16,11 @@ internal interface IDeployment
         /// </summary>
         string ProjectName { get; }
 
+        /// <summary>
+        /// Returns the project root. The location of Pulumi.yaml.
+        /// </summary>
+        string? RootDirectory { get; }
```
Frassle, line 22:
```suggestion
        string RootDirectory { get; }
```

sdk/Pulumi/Deployment/DeploymentInstance.cs
```diff
@@ -24,6 +24,11 @@ internal DeploymentInstance(IDeployment deployment)
         /// </summary>
         public string ProjectName => _deployment.ProjectName;
 
+        /// <summary>
+        /// Root directory. Defaults to <b>"rootDirectory"</b> if not specified.
```
Frassle, line 28:
```suggestion
        /// Root directory, that is the location of the Pulumi.yaml file.
```

sdk/Pulumi/Deployment/Deployment.cs
```diff
@@ -169,6 +173,19 @@ internal Deployment(IEngine engine, IMonitor monitor, TestOptions? options)
 
         string IDeployment.OrganizationName => _organizationName;
         string IDeployment.ProjectName => _projectName;
+        string IDeployment.RootDirectory
+        {
+            get
+            {
+              if (_rootDirectory == null)
+              {
+                throw new InvalidOperationException("Root directory is not available in your version of the pulumi CLI, please upgrade to the latest version");
```
Frassle, line 182:
```suggestion
                throw new InvalidOperationException("The Pulumi CLI version does not support root directory. Please update the Pulumi CLI.");
```
Standard phrasing copied from other similar errors.

Frassle (approved):

PR #492, opened 2025-02-20
Update renovate config for submodules

brandonpollack23:
Added an empty config file in case we need overrides later for renovate.

Set up submodule update branch.


.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "pulumi"]
 	path = pulumi
 	url = https://github.com/pulumi/pulumi.git
+  branch = master
```
Frassle, line 4:
```suggestion
	branch = master
```

renovate.json5
Frassle, line 1: Does a file here do anything? I'd rather add this when we need it unless a file marker is important to semantics somehow.

Frassle: To be consistent we'd have these everywhere. I'm not sure we need that given we're all going to know that renovate is how we do upgrades.

Frassle: Looks like we've merged these everywhere else so consistency demands it goes here as well

Frassle (approved):

PR #497, opened 2025-02-21
Add a `make changelog` command for consistency with `pulumi/pulumi`

i-am-tom:
It wasn't immediately obvious to me how to produce a changelog, but `make changelog` was the first thing I tried. This PR adds a `make changelog` command to mimic `pulumi/pulumi`.


Frassle (approved):
I think we'll probably try and get the "changie new" everywhere eventually, but this helps in the meantime.

PR #498, opened 2025-02-21
Schema Analyzer to infer component schemas from classes

mikhailshilkov:
Implements `ComponentAnalyzer.GenerateSchema` which takes either an assembly or a list of component resources and produces a serializable package schema structure. Infers component resources and their dependent types. Here are the features that this PR implements and tests:

- [x] Inputs
- [x] Outputs
- [x] Plain types
- [x] Boolean, integer, number, string
- [x] Complex types
- [x] Nested complex types
- [x] Self-recursive complex types
- [x] Mutually recursive complex types
- [x] Required properties & inputs
- [ ] Component description
- [ ] Complex type description
- [ ] Property description
- [x] Optional
- [x] Array
- [x] Map
- [x] Asset/Archive
- [ ] Resource reference
- [ ] Discriminated Union
- [ ] Enum
- [ ] Methods
- [ ] References to component in same provider
- [ ] pulumi.Any

Resolve https://github.com/pulumi/pulumi-dotnet/issues/468 (we'll track missing features in separate issues)

### DIFF
--- a/sdk/Pulumi/Provider/ComponentAnalyzer.cs
+++ b/sdk/Pulumi/Provider/ComponentAnalyzer.cs
@@ -1,0 +1,393 @@
+// Copyright 2025, Pulumi Corporation
+
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using System.Reflection;
+namespace Pulumi.Experimental.Provider
+{
+    /// <summary>
+    /// Analyzes component resource types and generates a package schema.
+    /// </summary>
+    public sealed class ComponentAnalyzer
+    {
+        private readonly Metadata metadata;
+        private readonly Dictionary<string, ComplexTypeSpec> typeDefinitions = new();
+
+        private ComponentAnalyzer(Metadata metadata)
+        {
+            this.metadata = metadata;
+        }
+
+        /// <summary>
+        /// Analyzes the components in the given assembly and generates a package schema.
+        /// </summary>
+        /// <param name="metadata">The package metadata including name (required), version and display name (optional)</param>
+        /// <param name="assembly">The assembly containing component resource types to analyze</param>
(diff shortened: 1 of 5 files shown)


Frassle (commented):
Why does this just return a schema? Don't we need to keep the schema together with construct functions that can implement it

mikhailshilkov:
> Why does this just return a schema? Don't we need to keep the schema together with construct functions that can implement it

This is mostly a utility class, it will be used in a provider implementation similar to [this Java one](https://github.com/pulumi/pulumi-java/blob/main/sdk/java/pulumi/src/main/java/com/pulumi/provider/internal/ComponentProvider.java)

sdk/Pulumi/Provider/ComponentAnalyzer.cs
```diff
@@ -0,0 +1,393 @@
...
+
+        /// <summary>
+        /// Analyzes the specified component types and generates a package schema.
+        /// </summary>
+        /// <param name="metadata">The package metadata including name (required), version and display name (optional)</param>
+        /// <param name="componentTypes">The component resource types to analyze</param>
+        /// <returns>A PackageSpec containing the complete schema for all components and their types</returns>
+        public static PackageSpec GenerateSchema(Metadata metadata, params Type[] componentTypes)
+        {
+            if (metadata?.Name == null || string.IsNullOrWhiteSpace(metadata.Name))
```
tgummerer, line 44: Should we evaulate this against the same regex we have in [`pulumi.json`](https://github.com/pulumi/pulumi/blob/2348c62e6dcfbbaf737b67963546aa5110057d67/pkg/codegen/schema/pulumi.json#L11)?

Also micronit, I'd prefer curly braces in all if statements even if they are just one line.  Not sure we have a linter or guidelines for that though.

tgummerer (approved):

PR #501, opened 2025-02-21
Update to Go 1.24

Frassle:


pulumi-language-dotnet/main_test.go
```diff
@@ -182,7 +182,6 @@ func TestDeterminePackageDependency(t *testing.T) {
 }
 
 func TestBuildDll(t *testing.T) {
-	t.Parallel()
```
Frassle, line 185: Turns out this was racy, I kept hitting errors in conformance tests because the chdir below was happening while the conformance test was running.

justinvp: I guess we're not linting our go code in this repo? Might be worth a `//nolint:paralleltest // mutates cwd` comment so we don't mistakenly add back `t.Parallel()`.

justinvp (approved):

pulumi-language-dotnet/main_test.go
Frassle: Yeh will do. We should work out getting the linter running as well at some point.

PR #502, opened 2025-02-22
[sdk] Support parameterization for remote component resources

### LINKED ISSUE #503: [sdk] Remote component resources do not support parameterization
`ComponentResource`s do not accept a `RegisterPackageRequest` parameter in their constructor and as such do not support parameterized packages leading to runtime issues mentioning missing plugins:
```
error: Could not automatically download and install resource plugin 'pulumi-resource-<parameterized-package>' at version v0.0.x
```
We need to 
 - add a new constructor that accepts `RegisterPackageRequest`
 - implement sdk-gen support to emit the new parameter when a resource is marked `isComponent: true`

Zaid-Ajaj:
### Description

When generating parameterized .NET SDKs, resources that marked `isComponent: true` in the schema do not support parameterization leading to runtime errors that mentioned missing plugins
```
error: Could not automatically download and install resource plugin 'pulumi-resource-<parameterized-package>' at version v0.0.x
```
This PR adds SDK support for parameterized _remote_ component resources. Still required SDK-gen support from pulumi/pulumi to emit the actual `RegisterPackageRequest` in the newly added constructor overload 

Fixes #503

### DIFF
--- a/.changes/unreleased/Improvements-502.yaml
+++ b/.changes/unreleased/Improvements-502.yaml
@@ -1,0 +1,6 @@
+component: sdk
+kind: Improvements
+body: Support parameterization for remote component resources
+time: 2025-02-22T08:54:11.922966+01:00
+custom:
+    PR: "502"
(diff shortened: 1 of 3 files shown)


Frassle (commented):
Can we add a component to the parameterised conformance test in pu/pu? 
l2-parameterized-resource

We should have an invoke in that test as well.

Frassle (approved):

PR #513, opened 2025-02-26
Keep the main projects on net6.0

Frassle:
_Follow-up to_: https://github.com/pulumi/pulumi-dotnet/pull/500 and https://github.com/pulumi/pulumi/pull/18619

While net6 isn't actually in support anymore, a hard migration to net8 is causing lots of downstream issues. So leave these projects targeting net6 for now, they'll still build fine with net8.


EronWright (approved):
Thank you!

PR #517, opened 2025-02-27
Update changelog for v3.75.2

Frassle:
https://github.com/pulumi/pulumi-dotnet/pull/516 didn't update the CHANGELOG.md file so didn't trigger the release.


justinvp:
Ah, sorry about that. I was running into changie errors running the `changie batch auto` command because it didn't like the format of the `.changes/unreleased/Bug Fixes-515.yaml` file. I had to rename it locally to get the command to work without it erroring.

PR #523, opened 2025-02-27
Update `pulumi` submodule to v3.153.0 tag

justinvp:
Update `pulumi` submodule to the v3.153.0 tag. This brings in two new conformance tests.

- `l1-builtin-project-root` support is being added in #466 
- `l2-component-property-deps` is blocked on .NET programgen support for `Call` (and #488)

Somewhat suspicious that a couple generated SDKs changed (I ran the conformance tests with `PULUMI_ACCEPT=true`). Are we not snapshot comparing those?


Frassle:
> Somewhat suspicious that a couple generated SDKs changed (I ran the conformance tests with PULUMI_ACCEPT=true). Are we not snapshot comparing those?

Should be. Maybe they were committed by mistake and hadn't actually been running before this submodule update?

Frassle (approved):

PR #525, opened 2025-03-04
Print a warning if an output is stringified

Frassle:
DotNet fix for https://github.com/pulumi/pulumi/issues/13518

This adds log warnings to warn when an attempt is made to convert an output to a string.

I'm inclined not to add this behind a flag as cases of this are nearly always user mistakes that should be fixed.


.changes/unreleased/Improvements-525.yaml
```diff
@@ -0,0 +1,6 @@
+component: sdk
+kind: Improvements
+body: Trying to convert outputs to strings now logs a warning
```
tgummerer, line 3:
```suggestion
body: Log a warning when trying to convert outputs to strings
```

sdk/Pulumi.Tests/Core/OutputTests.cs
```diff
@@ -961,8 +961,21 @@ public void OutputToStringThrowsWhenSet(string? variableValue)
                 var o1 = CreateOutput(0, isKnown: true);
                 Assert.Throws<InvalidOperationException>(() => o1.ToString());
 
+                // If we don't throw check that we log a warning, and then return the expected warning string.
```
tgummerer, line 964: I can't comment there, but was the Pulumi submodule updated intentionally in this PR?

tgummerer (approved):

sdk/Pulumi.Tests/Core/OutputTests.cs
Frassle: No I'll fix that.

PR #527, opened 2025-03-06
Fix conversion for nested Output<T>

algompluecker:


algompluecker:
I had to upgrade xunit for my test. Therefore, I had to do some fixes in other tests that produced warnings and did not build anymore.

Frassle (approved):

PR #533, opened 2025-03-08
Add a self reference to CallRequest and ConstructRequest

Frassle:
Part of https://github.com/pulumi/pulumi/issues/18823

This adds a new `ProviderResource` property to `CallRequest` and `ConstructRequest` that the runtime fills in with a `DependencyProviderResource` if the engine told us our URN and ID in `Configure`.


sdk/Pulumi/Provider/Provider.cs
```diff
@@ -363,13 +363,15 @@ public DeleteRequest(Urn urn, string id, ImmutableDictionary<string, PropertyVal
 
     public sealed class ConstructRequest
     {
+        public ProviderResource? Provider { get; init; }
```
Frassle, line 366: Arguably this should be non-nullable and throw with a message like "self provider references aren't supported update your pulumi CLI", that would match how we handle other cases like this that depend on engine support.

sdk/Pulumi/Provider/Provider.cs
```diff
@@ -611,6 +615,9 @@ class ResourceProviderService : ResourceProvider.ResourceProviderBase, IDisposab
         private Provider? implementation;
         private readonly string version;
         private string? engineAddress;
+        // More recent versions of the engine send URN and ID of the provider to `Configure`. With that we can construct
+        // a `DependencyProviderResource` to use in `Construct` and `Call` to refer to the provider itself.
+        private string? providerSelfReference;
```
Frassle, line 620: There's no value in sending a ProviderResource to any of the normal CRUD methods, but maybe we should send the URN for logging reasons? That would allow you to log a message to the provider resource, while performing a Create/Update for another resource.

PR #536, opened 2025-03-10
Handle logging arguments in provider's getEngineAddress

mikhailshilkov:
Passing a --logflow argument to the CLI will make it pass all its arguments to providers. This PR ports this [Node.js implementation](https://github.com/pulumi/pulumi/blob/05970ff66567eeed7723d7b6ee2c37d3b464f580/sdk/nodejs/provider/internals.ts#L30-L53) and [Java implementation](https://github.com/pulumi/pulumi-java/pull/1708) that discards known logging parameters from engine address calculation.

~This change now requires an address while previously it was optional. I believe this is a correct change, again in line with other SDKs, but let me know if there are legit cases when address is not passed.~

### DIFF
--- a/sdk/Pulumi/Provider/Provider.cs
+++ b/sdk/Pulumi/Provider/Provider.cs
@@ -542,6 +542,8 @@ namespace Pulumi.Experimental.Provider
             // maxRpcMessageSize raises the gRPC Max message size from `4194304` (4mb) to `419430400` (400mb)
             var maxRpcMessageSize = 400 * 1024 * 1024;
 
+            var engineAddress = GetEngineAddress(args);
+
             return Host.CreateDefaultBuilder()
                 .ConfigureWebHostDefaults(webBuilder =>
                 {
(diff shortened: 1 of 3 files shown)


tgummerer (approved):

sdk/Pulumi/Provider/Provider.cs
```diff
@@ -600,6 +602,41 @@ internal static Microsoft.Extensions.Hosting.IHost BuildHost(
...
+        {
+            var cleanArgs = new List<string>();
+
+            for (int i = 0; i < args.Length; i++)
+            {
+                var arg = args[i];
+
+                // Skip logging-related arguments
+                if (arg == "--logtostderr") continue;
+                if (arg.StartsWith("-v=")) continue;
```
Frassle, line 616: I think this can also be passed like "-v10" without an =

sdk/Pulumi/Provider/Provider.cs
```diff
@@ -600,6 +602,41 @@ internal static Microsoft.Extensions.Hosting.IHost BuildHost(
...
+        {
+            var cleanArgs = new List<string>();
+
+            for (int i = 0; i < args.Length; i++)
+            {
+                var arg = args[i];
+
+                // Skip logging-related arguments
+                if (arg == "--logtostderr") continue;
+                if (arg.StartsWith("-v")) continue;
```
justinvp, line 616: `StartsWith` does a culture-sensitive comparison. It would be better to use `arg.StartsWith("-v", StringComparison.Ordinal)` here.

We should enable some FxCop rules (or equivalent) to catch things like this.

justinvp: https://github.com/pulumi/pulumi-dotnet/pull/559

PR #539, opened 2025-03-11
Make Input.ToOutput non-nullable

Frassle:
The extension method to turn an input to an output had a very misleading nullability annotation:
```
        public static Output<T> ToOutput<T>(this Input<T>? input)
            => input ?? Output.Create(default(T)!);
```

This then propagated to other methods like the `Apply` extension methods, and JSON converters.

I've moved this extension method to an instance method on `Input<T>` and removed the nullability from it (well there's no way to make an instance method nullable).

This is a breaking change but I've had multiple users complain to me about this method. I think there's further work that could be done on improving the nullability annotations around outputs.


Frassle:
OK there's something to be done here for sure because the current state of things leads to nasty runtime bugs but it's going to be a pain to untangle. I'll close this off and think about what we can do here without breaking the world.

PR #547, opened 2025-03-18
dotnet Pack: improve error message when build fails

tgummerer:
Currently when the build in Pack fails, we just show the exit code and nothing else.  Improve that by showing stdout and stderr as well. Unfortunately the `errutil.ErrorWithStderr` helper is not enough here as dotnet prints errors to stdout, so let's show the stdout as well.


Frassle (approved):

pulumi-language-dotnet/main.go
```diff
@@ -1130,24 +1131,24 @@ func (host *dotnetLanguageHost) Pack(ctx context.Context, req *pulumirpc.PackReq
...
 			opts.dotnetExec, "build", "-c", "Release")
 		cmd.Dir = req.PackageDirectory
-		return cmd.Run()
+		return cmd.CombinedOutput()
 	}
 
-	if err := build(); err != nil {
-		return nil, fmt.Errorf("build error before pack: %w", err)
+	if out, err := build(); err != nil {
+		return nil, errutil.ErrorWithStderr(err, fmt.Sprintf("build error before pack.  stdout: ", string(out)))
```
Frassle, line 1151:
```suggestion
		return nil, errutil.ErrorWithStderr(err, fmt.Sprintf("build error before pack. stdout: ", string(out)))
```

PR #555, opened 2025-03-26
Infer namespaces in dotnet

tgummerer:
Key changes around the ComponentProviderHost:

- Introduces a new property `Namespace` to package metadata
- For assembly name `Foo.Bar.Buzz` or `Foo.Buzz` calculated package name as `buzz` and namespace as `foo`, for assembly name `Buzz` leaves namespace empty
- Namespace and package names are kebab-cased with Humanizer
- Changes the ComponentProvider constructor to accept the entire `metadata` record, which is built in the host
- ComponentProviderHost is the still "simple mode" entry point with sensible defaults, users can eject to ComponentProvider pretty easily

### DIFF
--- a/.changes/unreleased/Improvements-555.yaml
+++ b/.changes/unreleased/Improvements-555.yaml
@@ -1,0 +1,6 @@
+component: sdk/provider
+kind: Improvements
+body: Infer package name and namespace in ComponentProviderHost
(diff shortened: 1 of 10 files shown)


mikhailshilkov (approved):
Approving by verbal ok from @tgummerer

PR #561, opened 2025-03-28
Capture async context for engine callback methods

Frassle:
The `Callback` system is what allows the engine to callback to the language runtime to invoke generic code. Currently just used for the transforms system, but soon to be extended for other uses as well.

We were incorrectly not propagating the async context from where the callback system was created (somewhere inside the users program as part of resource registration call) to the threads that the callbacks run on (in the kestral server threadpool).

This meant that inside a transform function you couldn't access any of the async context that you had access to just outside the transform function. Importantly this includes the deployment instance that allows access to the engine.

We probably wouldn't recommend registering more resources from the transform function (although, you _could_) but users definitely should have access to the engine logging system from inside the transform function.

This PR adds the requisite ExecutionContext capture and run calls to copy and restore the async context from the main program to the threads running the callback functions.

### DIFF
--- a/sdk/Pulumi/Deployment/Deployment.cs
+++ b/sdk/Pulumi/Deployment/Deployment.cs
@@ -183,8 +183,10 @@ namespace Pulumi
                 return _callbacks;
             }
 
-            // Atomically allocate a callbacks instance to use
-            var callbacks = new CallbacksHost();
+            // Atomically allocate a callbacks instance to use, capture the current async context and pass it
+            // to the callbacks host. Importantly this captures the current Deployment.Instance, so basic
+            // things like engine logging work in transforms.
+            var callbacks = new CallbacksHost(System.Threading.ExecutionContext.Capture());
             var current = Interlocked.CompareExchange(ref _callbacks, callbacks, null);
             if (current == null)
             {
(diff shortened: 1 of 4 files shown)


sdk/Pulumi/Deployment/Deployment.cs
```diff
@@ -183,8 +183,10 @@ internal async Task<CallbacksHost> GetCallbacksAsync(CancellationToken cancellat
                 return _callbacks;
             }
 
-            // Atomically allocate a callbacks instance to use
-            var callbacks = new CallbacksHost();
+            // Atomically allocate a callbacks instance to use, capture the current async context and pass it
+            // to the callbacks host. Importantly this captures the current Deployment.Instance, so basic
+            // things like engine logging work in transforms.
+            var callbacks = new CallbacksHost(System.Threading.ExecutionContext.Capture());
```
justinvp, line 189: Nit: Aren't we already importing `System.Threading`? Is there some other reason it needs to be fully qualified?

```suggestion
            var callbacks = new CallbacksHost(ExecutionContext.Capture());
```

justinvp (approved):
LGTM

PR #564, opened 2025-04-01
Use a more aggressive set of analysis rules

justinvp:
The first commit bumps from `Default` to `Minimum`. The second commit bumps from `Minimum` to `Recommended`.

See https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#analysismode for more details.

### DIFF
--- a/sdk/Pulumi/Core/Input.cs
+++ b/sdk/Pulumi/Core/Input.cs
@@ -51,25 +51,33 @@ namespace Pulumi
         /// <summary>
         /// <see cref="Output{T}.Apply{U}(Func{T, Output{U}?})"/> for more details.
         /// </summary>
+#pragma warning disable CA1715 // Identifiers should have correct prefix
         public static Output<U> Apply<T, U>(this Input<T>? input, Func<T, U> func)
+#pragma warning restore CA1715 // Identifiers should have correct prefix
             => input.ToOutput().Apply(func);
 
         /// <summary>
         /// <see cref="Output{T}.Apply{U}(Func{T, Output{U}?})"/> for more details.
         /// </summary>
+#pragma warning disable CA1715 // Identifiers should have correct prefix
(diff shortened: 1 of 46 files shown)


sdk/Pulumi/Provider/Provider.cs
```diff
@@ -504,7 +504,12 @@ public static Task Serve(string[] args, string? version, Func<IHost, Provider> f
             return Serve(args, version, factory, cancellationToken, System.Console.Out);
         }
 
+// Consider getting rid of this overload or changing it so the CancellationToken is last.
```
Frassle, line 510: This is tagged experimental, just remove it.

sdk/Pulumi/Core/Input.cs
```diff
@@ -51,25 +51,33 @@ public static class InputExtensions
         /// <summary>
         /// <see cref="Output{T}.Apply{U}(Func{T, Output{U}?})"/> for more details.
         /// </summary>
+#pragma warning disable CA1715 // Identifiers should have correct prefix
```
Frassle, line 54: Not convinced these identifier suffix/prefix rules are worth having turned on.

justinvp: I was initially on the fence on these too, but I kind of want them on. I want to make sure we're naming Attributes and Exceptions with the right suffix (which is something I've seen previously in code reviews, that I luckily happened to catch).

I could be convinced to disable the prefix rule, since we tend to use `U` for generic type names rather than the preferred "prefix with `T`". But it does check that interfaces are prefixed with `I`, which I kind of want to keep. I wish I could easily ignore the generic type name one but keep the interface rule.

Frassle: Maybe we should just do a pass to change to T prefixes then, that seems to be the bulk of the warning disables.

Frassle (approved):

PR #567, opened 2025-04-05
Component-level debugging support (dotnet)

EronWright:
Extends debugging support to components via `RunPlugin`.

### DIFF
--- a/pulumi-language-dotnet/main.go
+++ b/pulumi-language-dotnet/main.go
@@ -740,7 +740,11 @@ func (host *dotnetLanguageHost) Run(ctx context.Context, req *pulumirpc.RunReque
 		ctx, cancel := context.WithCancel(ctx)
 		defer cancel()
 		go func() {
-			err = startDebugging(ctx, engineClient, cmd)
+			info := programInfo{
+				Name: fmt.Sprintf("%s (program)", req.Project),
+				Cwd:  cmd.Dir,
+			}
+			err = startDebugging(ctx, engineClient, cmd, info)
 			if err != nil {
 				// kill the program if we can't start debugging.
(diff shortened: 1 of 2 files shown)


pulumi-language-dotnet/main.go
```diff
@@ -1064,8 +1075,31 @@ func (host *dotnetLanguageHost) RunPlugin(
 			return err
 		}
 
-		// Now run from source without re-building.
-		args = append(args, "run", "--no-build", "--project", project, "--")
```
EronWright, line 1068: Here's a brief explanation for what's happening here, which is a small change to how the pre-built executable is run. 

As before, the plugin builds the project before running it, to capture build output for the display purposes. Then it runs the pre-built executable. Previously, we used `dotnet run --no-build --project ...` to run the executable. The `run` command automatically locates and executes the pre-built executable (e.g. `<project>/bin/Debug/net8.0/dotnet-components.dll`).

The issue with that approach is that runs the executable as a sub-process that is impractical to attach to. A simpler approach is to run the executable using `dotnet <assembly-name>`, which one may attach to directly and is how the Run RPC is implemented.

To use `dotnet <assembly-name>`, one must locate the assembly. The `Run` code uses an awkward approach, and I would like to try a different way that is based on interrogating msbuild for information about the build outputs:

```jsonc
// dotnet msbuild -getProperty:OutputPath -getProperty:AssemblyName
{
  "Properties": {
    "OutputPath": "bin/Debug/net8.0/",
    "AssemblyName": "dotnet-components"
  }
}
```

_I do think that `dotnet msbuild -getProperty` is a portable command; anyone disagree? ([ref](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-msbuild))_

PR #591, opened 2025-05-08
Update PropertyValue to match the new Go and Python SDKs

Frassle:
This removes `OutputReference` and the nested `PropertyValue` to indicate secrets. Instead every `PropertyValue` tracks secretness and dependencies.

### DIFF
--- a/sdk/Pulumi/Provider/PropertyValue.cs
+++ b/sdk/Pulumi/Provider/PropertyValue.cs
@@ -24,9 +24,7 @@ namespace Pulumi.Experimental.Provider
 #pragma warning restore CA1720 // Identifier contains type name
         Asset,
         Archive,
-        Secret,
         Resource,
-        Output,
         Computed,
     }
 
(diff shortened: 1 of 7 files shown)


sdk/Pulumi/Provider/PropertyValue.cs
```diff
@@ -198,26 +120,6 @@ public PropertyValueType Type
             }
         }
 
-        // Unwraps any outer secret or output values.
-        public PropertyValue Unwrap()
```
iwahbe, line 202: ❤️

iwahbe (approved):

iwahbe:
One unrelated nit, I noticed that we call the key->value `PropertyValue` `Object` in C#. In Go (`property.Value`), we call it `Map`.

Frassle:
Good catch, I'll rename that and make sure python is Map as well

julienp (approved):

PR #597, opened 2025-05-13
Use repo-specific ESC environment

pgavlin:
This repository has repository-specific secrets that need to be migrated to ESC. These changes replace the usage of the common GHA ESC environment with the repository-specific environment.


komalali (approved):

justinvp:
Do we need it for this repo?

- `CODECOV_TOKEN` - I think we can get rid of this, given this banner at https://app.codecov.io/gh/pulumi/pulumi-dotnet:

  Screenshot 2025-05-13 at 4 07 25 PM

- `TMP_GIST_TOKEN` - Doesn't look used. I think we can delete it.

tgummerer:
> * `TMP_GIST_TOKEN` - Doesn't look used. I think we can delete it.

Yep, I added that at some point for some testing, we can indeed delete that (and I did so now).

PR #600, opened 2025-05-15
implement Plugin debugging for dotnet

tgummerer:
Implement plugin debugging for dotnet.  This is unfortunately quite awkward, because dotnet doesn't have a command line option to say start the program and attach the program. Instead the program itself needs to check whether a debugger is attached or not, and wait for it otherwise.

This PR implements this the same way we do for programs, by setting an environment variable and then waiting in the provider startup code. Unfortunately this means it only works for providers written with our provider framework currently, but not with other types of plugins.

/xref https://github.com/pulumi/home/issues/4061


tgummerer:
I *think* CI here is failing because we need a pulumi version that includes https://github.com/pulumi/pulumi/pull/19532 to be able to pass the `--attach-debugger` flag correctly. This also requires https://github.com/pulumi/pulumi-dotnet/pull/604 to be merged in addition to the pulumi/pulumi PR.

sdk/Pulumi/Provider/Provider.cs
```diff
@@ -504,6 +504,15 @@ public virtual Task<CallResponse> Call(CallRequest request, CancellationToken ct
 
         public static async Task Serve(string[] args, string? version, Func<IHost, Provider> factory, System.Threading.CancellationToken cancellationToken)
         {
+            var value = System.Environment.GetEnvironmentVariable("PULUMI_ATTACH_DEBUGGER");
```
Frassle, line 507: I know this is just a continuation of the way we do this for programs, but urgh it feels grim.

Frassle (approved):

sdk/Pulumi/Provider/Provider.cs
tgummerer: Yeah, it's kinda sad, but unfortunately I do not know a better way.

PR #605, opened 2025-05-19
Warn on constructor argument mismatches in `ConvertObject`

i-am-tom:
Before this PR, converting between two objects whose arguments did not match would fail silently if the values could be casted (or similar). This makes it harder to track down mismatches. This PR introduces a warning when this happens.

### DIFF
--- a/sdk/Pulumi/Serialization/Converter.cs
+++ b/sdk/Pulumi/Serialization/Converter.cs
@@ -201,7 +201,7 @@ namespace Pulumi.Serialization
 
             var constructorParameters = constructor.GetParameters();
             var arguments = new object?[constructorParameters.Length];
-            var warning = "";
+            var warning = null;
 
             for (int i = 0, n = constructorParameters.Length; i < n; i++)
             {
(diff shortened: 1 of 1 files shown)


sdk/Pulumi/Serialization/Converter.cs
```diff
@@ -201,11 +201,13 @@ private static (object?, string?) TryConvertObject(Action<string> warn, string c
 
             var constructorParameters = constructor.GetParameters();
             var arguments = new object?[constructorParameters.Length];
+            var warning = "";
```
tgummerer, line 204: Should this be `var warning = null;` to preserve previous behaviour?

tgummerer (commented):
This makes sense to me (obviously since we paired on this :D), but I'll let someone with more C# knowledge approve. 

Also we should probably have a test for this showing combinations where we warn and don't warn.

Frassle:
> would fail silently if the values could be casted (or similar)

If the values could cast it failed? What if they couldn't cast? This definitely needs some demonstration tests.

PR #608, opened 2025-05-20
Add dotnet sdk changes to support scalar method call returns

brandonpollack23:
Much like pulumi/pulumi#19387, this enables dotnet scalar return values from method calls,
but now in dotnet.

Part of pulumi/pulumi#7345


sdk/Pulumi/Deployment/DeploymentInstance.cs
```diff
@@ -155,6 +157,49 @@ public Output<T> Call<T>(
...
+                            "expected on object with a single member, but had " + properties.Length
+                        );
+                    }
+                    PropertyInfo singleProperty = properties[0];
+                    if (!singleProperty.CanRead)
+                    {
+                        throw new InvalidOperationException("expected property to be readable");
+                    }
+
+                    object? value = singleProperty.GetValue(obj, null);
```
Frassle, line 190: We might need to re-cast here. Consider CallSingle<int> but the wire format only has floats so obj at this point will be a `double` not an `int`, but for a normal property I think we have some cast logic to do that conversion.

Frassle: The wire format, and so the default deserialiser to object only handles floats, not integers. So a number on the wire format always deserialises to `double` unless there's extra typing information to guide it. Which there won't be for the Call here because we're doing a `Call<object>`. But the property might be intended to be an `int` and will fail the type check here because it will actually be a `double` from the Call.

brandonpollack23: I think this change handles it let me know

Frassle: I mean yes for doubles (although I think it's overkill, we only ever have int or double as a target type), but there's also enums to handle. See the code for TryConvertObject in sdk/Pulumi/Serialization/Converter.cs.

brandonpollack23: this should do that

PR #615, opened 2025-05-26
dont set the stack as the parent if urn is set

jkerken:
If an engine read is to be performed as the result of deserializing a Output<SomeResourceType> it throws:

InvalidOperationException("Trying to acquire Deployment.Stack before 'Run' was called.")

Since inside of MLC Construct no Stack is set and everything needs to have a parent.

The parent is only neede for transformation apply, so I think it safe that just set this to null here?

I am not sure about the compution and applying of transformations etc that could all be skiped, but the options that created from Pulumi.ResourcePackages.TryConstruct anyways dont set anything other than URN so we can also just leave it as is


Frassle:
This looks reasonable to me but it would be good to have a test showing that it fixes the issue mentioned (a resource read in an MLC). I imagine it's possible to add something to the test provider to test this.

.changes/unreleased/bug-fixes-615.yaml
```diff
@@ -0,0 +1,6 @@
+component: sdk
+kind: bug-fixes
+body: disable stack auto parenting for resources that are read from the engine
```
tgummerer, line 3:
```suggestion
body: Disable stack auto parenting for resources that are read from the engine
```

Frassle (approved):

PR #619, opened 2025-05-29
Move Provider.Urn to the core SDK

Frassle:
We need to start splitting the "Provider" namespace up to also support policy. One part that will be used by both is the concept of "URN". This is stable enough that I think we're fine just moving it down into the core SDK replacing the URN class currently there with a value struct.

### DIFF
--- a/sdk/Pulumi/Core/Urn.cs
+++ b/sdk/Pulumi/Core/Urn.cs
@@ -11,7 +11,7 @@ namespace Pulumi
     /// </summary>
     public readonly struct Urn : IEquatable<Urn>
     {
-        readonly string _value;
+        private readonly string _value;
 
         public Urn(string value)
         {
@@ -31,52 +31,38 @@ namespace Pulumi
         /// <summary>
         /// Returns the string representation of this URN.
         /// </summary>
-        override public string ToString() => _value;
+        public override string ToString() => _value;
 
         /// <summary>
         /// Determines whether the specified object is equal to the current URN.
         /// </summary>
         public override bool Equals([NotNullWhen(true)] object? obj)
-        {
-            if (obj is Urn other)
-            {
-                return Equals(this, other);
-            }
-            return false;
-        }
+            => obj is Urn other && Equals(other);
 
         /// <summary>
         /// Determines whether the specified URN is equal to the current URN.
         /// </summary>
         public bool Equals(Urn other)
-        {
-            return string.Equals(_value, other._value, StringComparison.Ordinal);
-        }
+            => string.Equals(_value, other._value, StringComparison.Ordinal);
 
         /// <summary>
         /// Returns a hash code for this URN.
         /// </summary>
-        override public int GetHashCode()
-        {
-            return _value.GetHashCode(StringComparison.Ordinal);
-        }
+        public override int GetHashCode()
+            => _value.GetHashCode(StringComparison.Ordinal);
 
 
         /// <summary>
         /// Determines whether two URN values are equal.
         /// </summary>
         public static bool operator ==(Urn left, Urn right)
-        {
-            return left.Equals(right);
-        }
+            => left.Equals(right);
 
         /// <summary>
         /// Determines whether two URN values are not equal.
         /// </summary>
         public static bool operator !=(Urn left, Urn right)
-        {
-            return !(left == right);
-        }
+            => !(left == right);
 
         /// <summary>
         /// Computes a URN from the combination of a resource name, resource type, optional parent,


sdk/Pulumi/Core/Urn.cs
```diff
@@ -1,15 +1,83 @@
...
 namespace Pulumi
 {
     /// <summary>
     /// An automatically generated logical URN, used to stably identify resources. These are created
     /// automatically by Pulumi to identify resources.  They cannot be manually constructed.
     /// </summary>
-    public static class Urn
+    public readonly struct Urn : IEquatable<Urn>
     {
+        readonly string _value;
```
justinvp, line 14:
```suggestion
        private readonly string _value;
```

sdk/Pulumi/Core/Urn.cs
```diff
@@ -1,15 +1,83 @@
...
+
+        /// <summary>
+        /// Implicitly converts a URN to a string.
+        /// </summary>
+        public static implicit operator string(Urn value) => value._value;
+
+        /// <summary>
+        /// Returns the string representation of this URN.
+        /// </summary>
+        override public string ToString() => _value;
```
justinvp, line 34:
```suggestion
        public override string ToString() => _value;
```

sdk/Pulumi/Core/Urn.cs
```diff
@@ -1,15 +1,83 @@
...
+        /// </summary>
+        public bool Equals(Urn other)
+        {
+            return string.Equals(_value, other._value, StringComparison.Ordinal);
+        }
+
+        /// <summary>
+        /// Returns a hash code for this URN.
+        /// </summary>
+        override public int GetHashCode()
```
justinvp, line 59:
```suggestion
        public override int GetHashCode()
```

Zaid-Ajaj (approved):

sdk/Pulumi/Core/Urn.cs
```diff
@@ -1,15 +1,83 @@
...
+        /// Determines whether the specified object is equal to the current URN.
+        /// </summary>
+        public override bool Equals([NotNullWhen(true)] object? obj)
+        {
+            if (obj is Urn other)
+            {
+                return Equals(this, other);
+            }
+            return false;
+        }
```
justinvp, line 46: `Equals(this, other)` is not the `Equals` you want; you want `Equals(other)`. These lines could be simplified as:

```suggestion
            => obj is Urn other && Equals(other);
```

sdk/Pulumi/Core/Urn.cs
```diff
@@ -1,15 +1,83 @@
...
+
+        /// <summary>
+        /// Determines whether the specified URN is equal to the current URN.
+        /// </summary>
+        public bool Equals(Urn other)
+        {
+            return string.Equals(_value, other._value, StringComparison.Ordinal);
+        }
```
justinvp, line 54:
```suggestion
            => string.Equals(_value, other._value, StringComparison.Ordinal);
```

sdk/Pulumi/Core/Urn.cs
```diff
@@ -1,15 +1,83 @@
...
+
+        /// <summary>
+        /// Returns a hash code for this URN.
+        /// </summary>
+        override public int GetHashCode()
+        {
+            return _value.GetHashCode(StringComparison.Ordinal);
+        }
```
justinvp, line 62:
```suggestion
            => _value.GetHashCode(StringComparison.Ordinal);
```

sdk/Pulumi/Core/Urn.cs
```diff
@@ -1,15 +1,83 @@
...
+
+
+        /// <summary>
+        /// Determines whether two URN values are equal.
+        /// </summary>
+        public static bool operator ==(Urn left, Urn right)
+        {
+            return left.Equals(right);
+        }
```
justinvp, line 71:
```suggestion
            => left.Equals(right);
```

sdk/Pulumi/Core/Urn.cs
```diff
@@ -1,15 +1,83 @@
...
+
+        /// <summary>
+        /// Determines whether two URN values are not equal.
+        /// </summary>
+        public static bool operator !=(Urn left, Urn right)
+        {
+            return !(left == right);
+        }
```
justinvp, line 79:
```suggestion
            => !(left == right);
```

justinvp (approved):
One comment on Equals, the rest are nits. LGTM otherwise.

PR #629, opened 2025-06-11
`Makefile` migration: `make clean`

i-am-tom:
A conversation I had with @lunaris recently: our repositories should have common commands for common actions (code quality, changelog, release, etc). Currently, `pulumi-dotnet` has an awkward split between `Makefile` and `build/Project.fs`, and this PR is the start of an attempt to bring everything possible over to the `Makefile`. I'll leave pass-through calls in the `Project.fs` file for now so all previous commands will still work.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,3 @@
-# Clean all generated SDK files
-clean::
-	cd sdk && dotnet clean
-	rm -rf sdk/*/{bin,obj}
-
 GO := go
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -1,3 +1,8 @@
+# Clean all generated SDK files
```
tgummerer, line 1: Why only the `SDK` files?  Should it not clean the other build artifacts as well?

tgummerer (approved):

PR #635, opened 2025-06-11
fix release workflow

tgummerer:
The release workflow for the latest release has failed, because goreleaser got updated to 2.x, and --skip-validate has been removed in that version (see https://goreleaser.com/deprecations/#-skip).

We could pin the goreleaser version, but we might as well just update it since it seems a straightforward fix.

Fixes https://github.com/pulumi/pulumi-dotnet/issues/632

### DIFF
--- a/.changes/unreleased/bug-fixes-635.yaml
+++ b/.changes/unreleased/bug-fixes-635.yaml
@@ -1,0 +1,6 @@
+component: build
+kind: bug-fixes
+body: Fix release process
+time: 2025-06-11T13:15:03.188755448+02:00
+custom:
+    PR: "635"


.changes/unreleased/bug-fixes-635.yaml
```diff
@@ -0,0 +1,6 @@
+component: build
+kind: bug-fixes
+body: Fix release process
```
tgummerer, line 3: This probably doesn't deserve a changelog, but it's going to be useful to release a .1 release that hopefully succeeds.

justinvp (approved):

PR #645, opened 2025-06-12
Start of policy support

Frassle:
This is not complete, but it's enough to get the first two conformance tests we wrote for policy passing.

This is pretty much a copy of the Go SDK (which heavily copied from the python and nodejs SDKs), plus taking some ideas from the C# provider SDK.

A few things worth calling out.

1. I've added a nuget cache clear to the start of the language tests. Me and Brandon both wasted loads of time when dealing with dotnet conformance tests because nuget was picking up old builds of the core SDK from caches. No more, we know nuke the cache before running the tests.
2. I had to add `WaitDelay` to the run plugin code. Seemed the dotnet run of the analyzer plugin could leave stdio pipes open even though it had terminated. Without `WaitDelay` run plugin would block forever on `cmd.Wait()`.
3. Some parts of the policy SDK are partly filled in (Like config schema and properties) but not working and not currently tested. I think they're about the right shape, but we might tweak them once we have conformance tests in place to check them.


sdk/Pulumi/Policy/Manager.cs
```diff
@@ -0,0 +1,44 @@
...
+
+namespace Pulumi.Experimental.Policy
+{
+    /// <summary>
+    /// An interface into the policy system for reporting policy violations.
+    /// </summary>
+    public interface IPolicyManager
+    {
+        /// <summary>
+        /// ReportViolation reports a policy violation with the given message and optional URN (it can be left null or empty).
```
rosieks, line 25: ReportViolation is redundant. There no reason to write method name if summary.

PR #653, opened 2025-06-18
Don’t remove bin,obj from the root during clean

julienp:
This holds the build and deps of the scripts in `Program.fs`, don’t delete this, as we might be calling `make clean` in the middle of some script.

The FSharp implementation prior to https://github.com/pulumi/pulumi-dotnet/pull/629 did not remove these files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ changelog::
 
 clean:
 	cd sdk && dotnet clean
-	rm -rf {bin,obj} sdk/*/{bin,obj}
+	rm -rf sdk/*/{bin,obj}
 
 test_integration:: build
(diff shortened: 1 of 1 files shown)


julienp:
The problem here is that when you run `dotnet run publish-sdks`, dotnet will compile the `Build.fsproj` project, using `./obj` and `./bin` to do so. This is the project that holds the `publish-sdks` script, and others. While this script executes, it calls `make clean`, which deletes its own compilation artefacts. Presumably it depends on the OS and lazy loading etc what happens next.

Prior to https://github.com/pulumi/pulumi-dotnet/pull/629 the equivalent FSharp implementation of `make clean` did not remove these files, only the ones inside `sdk/*`.

PR #655, opened 2025-06-20
`Makefile` migration: `make lint`/`make lint_fix`/`make format`/`make format_check`

i-am-tom:
After some discussion on Slack, here's where we are:

- `make format` will apply formatting changes and write to disk.
- `make lint` will report errors (including formatting errors) without writing to disk.
- `make format_check` will report formatting errors without writing to disk.
- `make lint_fix` will apply linting changes and write to disk.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -51,6 +51,12 @@ jobs:
         with:
           # Install 8.0 for Build.fsproj
           dotnet-version: 8.0.x
+      - name: Set up Go 1.24.x
+        uses: actions/setup-go@v5
+        with:
+          go-version: 1.24.x
+      - name: Install `gofumpt`
+        run: go install mvdan.cc/gofumpt@latest
       - name: Run `make format_check`
         run: make format_check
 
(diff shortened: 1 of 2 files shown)


julienp:
Looks mostly good!

There's an issue in the format action: https://github.com/pulumi/pulumi-dotnet/actions/runs/15780786287/job/44485672179?pr=655#step:6:18

```
cd sdk && dotnet format --verify-no-changes
Warnings were encountered while loading the workspace. Set the verbosity option to the 'diagnostic' level to log warnings.
/bin/sh: 1: gofumpt: not found
```

Somewhat annoying that this doesn't fail the job

i-am-tom:
> Probably need to review the jobs to see if there's any other cases.

@julienp Looks like the problem was that the `make` subshells don't propagate errors up to the command. Have fixed this

i-am-tom:
> Do we even need the format check as a separate target?

@tgummerer I don't know, but I'd be inclined to say it's a conversation for beyond this PR - I'm just trying to get rid of the F# code for now, then we can figure out whether it's the best way to do everything

PR #662, opened 2025-06-27
`Makefile` migration: `make all`, `make help`, `make changelog`, `make test_fast`, remove `restore` calls

i-am-tom:
A few little changes to get more in line with the `Makefile` proposal.

Personally, the further we get into this, the more it feels to me that something like [casey/just](https://github.com/casey/just) would be a far better fit for what we want. Even `mise` has basic task-running capabilities. Maintaining a `help` command with a useful amount of documentation using `make` is a big effort, and the compilers we use are advanced enough that we probably never want a non-`PHONY` target in this file. We only really need a task runner.


Makefile
```diff
@@ -111,12 +118,10 @@ test_integration: build clean
 
 .PHONY: test_sdk
 test_sdk: build clean
```
tgummerer, line 120: If we don't restore, should we still clean?

Makefile
```diff
@@ -126,16 +131,20 @@ test_coverage: test_sdk_coverage test_sdk_automation_coverage
...
 			-p:PulumiSdkVersion=$(SDK_VERSION) \
 			-p:CollectCoverage=true \
 			-p:CoverletOutputFormat=cobertura \
 			-p:CoverletOutput=./coverage/coverage.pulumi.automation.xml
+
+.PHONY: help
+help:
+	@awk 'BEGIN { print "Available targets:" } \
+		/^##/ { help=$$0; next } \
+		/^[a-zA-Z0-9_-]+:/ { sub(/:.*/, "", $$1); printf "  %-20s %s\n", $$1, help; help="" }' $(MAKEFILE_LIST)
```
tgummerer, line 150: I'm just going to trust that this works :sweat_smile:

tgummerer (approved):

Issue #672, opened 2025-07-15
Lifecycle hooks not running through stack transforms


pierskarsenbarg:
### What happened?

Using the following as a Pulumi program (coming from [here](https://github.com/pierskarsenbarg/lifecycle-hook-issues/blob/main/Program.cs)):

```csharp
using Pulumi;
using System;
using System.Threading.Tasks;
using System.Threading;
using Pulumi.Random;

class Program
{
    static Task<int> Main() => Deployment.RunAsync<MyStack>();
}

class MyStack : Stack
{
    private static ResourceHook beforeHook()
    {
        return new ResourceHook("before", async (args, cancellationToken) =>
        {
            Log.Info($"Resource {args.Urn} create started at {DateTime.Now.ToString()}");
            return;
        });
    }

    private static ResourceHook afterHook()
    {
        return new ResourceHook("after", async (args, cancellationToken) =>
        {
            Log.Info($"Resource {args.Id} create ended at {DateTime.Now.ToString()}");
            return;
        });
    }

    private static async Task<ResourceTransformResult?> StackTransform(ResourceTransformArgs args, CancellationToken ct)
    {
        if (args.Type != "pulumi:pulumi:Stack")
        {
            Log.Info($"Running transform on {args.Name} {args.Type}");

            var resourceArgs = args.Args;
            resourceArgs = resourceArgs.SetItem("length", "3");

            var options = CustomResourceOptions.Merge(
                (CustomResourceOptions)args.Options,
                new CustomResourceOptions
                {
                    Hooks =
                    {
                        BeforeCreate = [beforeHook()],
                        AfterCreate = [afterHook()]
                    },
                    AdditionalSecretOutputs = ["length"]
                }
            );
            return new ResourceTransformResult(resourceArgs, options);
        }
        return null;
    }

    public MyStack() : base(new StackOptions { ResourceTransforms = { StackTransform } })
    {
        var pet = new RandomPet("pet", new()
        {
            Length = 2
        });

        this.PetName = pet.Id.Apply(m => m);
    }

    [Output]
    public Output<string> PetName { get; private set; }
}
```

The length of the pet name is changed and the `Running transform` [log line](https://github.com/pierskarsenbarg/lifecycle-hook-issues/blob/e17956c1ed1227655c57523761f67d2855c7972c/Program.cs#L36) can be seen in the output, but the log lines from the hooks aren't seen.

### Example

[...]

PR #675, opened 2025-07-17
Allow setting resource hooks in transforms

julienp:
Similar to https://github.com/pulumi/pulumi/pull/20052 & https://github.com/pulumi/pulumi/pull/20057, respect resource hooks set in transforms.

Fixes https://github.com/pulumi/pulumi-dotnet/issues/672

### DIFF
--- a/integration_tests/transform_resource_hooks/Program.cs
+++ b/integration_tests/transform_resource_hooks/Program.cs
@@ -1,0 +1,64 @@
+// Copyright 2025, Pulumi Corporation.  All rights reserved.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Pulumi;
+
+class TransformResourceHooksStack : Stack
+{
+    public TransformResourceHooksStack()
+    {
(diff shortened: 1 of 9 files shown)


integration_tests/transform_resource_hooks/Program.cs
```diff
@@ -0,0 +1,66 @@
...
+                throw new InvalidOperationException($"Expected type 'testprovider:index:Random', got {resourceType}");
+            }
+            Console.WriteLine($"Hook was called with length = {args.NewInputs?["length"]}");
+            return Task.CompletedTask;
+        }
+
+        static Task<ResourceTransformResult?> Transform(ResourceTransformArgs args, CancellationToken cancellationToken = default)
+        {
+            var opts = args.Options;
+            var newHooks = opts.Hooks.Clone();
```
Frassle, line 27: Do we need to clone? Can't we just mutate the opts directly?

Frassle (approved):

Issue #686, opened 2025-08-19
Program does not appear to reference any 'Pulumi.*' packages


johnkors:
### What happened?

I can't get the pulumi cli to run on any of my .NET projects. Any hints? I've tried downgrading from 3.189 to previous versions, but that did not help.

```
➜  infra git:(main) ✗ pulumi version
v3.189.0
```

```
➜  Infra git:(main) ✗ pulumi up
Previewing update (dev)

View in Browser (Ctrl+O): https://app.pulumi.com/omit/omit/dev/previews/…

     Type                 Name            Plan     
     pulumi:pulumi:Stack  omitted-dev           

error: failed to discover package requirements: unexpected output from 'dotnet list package --include-transitive'. Program does not appear to reference any 'Pulumi.*' packages
```

```
➜  Infra git:(main) ✗ dotnet list package --include-transitive
Project 'Infra' has the following package references
   [net9.0]: 
   Top-level Package         Requested   Resolved
   > Pulumi                  3.87.0      3.87.0  
   > Pulumi.AzureNative      3.7.1       3.7.1   

```

### Example

I have no idea how to resolve this, totally blocked.

### Output of `pulumi about`

```
➜  Infra git:(main) ✗ pulumi about
running 'dotnet build -nologo'
  Determining projects to restore...

  All projects are up-to-date for restore.

  Infra -> /Users/johnkors/kode/… omitted





Build succeeded.

    0 Warning(s)
    0 Error(s)


Time Elapsed 00:00:00.49

'dotnet build -nologo' completed successfully
CLI          
Version      3.190.0
Go Version   go1.25.0
Go Compiler  gc

Host     
OS       darwin
Version  15.6
Arch     arm64

This project is written in dotnet: executable='/usr/local/share/dotnet/dotnet' version='9.0.303'

Current Stack: omitted/deltapower/yolo

Found no resources associated with yolo

Found no pending operations associated with yolo

Backend        
Name           pulumi.com
URL            https://app.pulumi.com/omitted
User           omitted
Organizations  johnkors, omitted
Token type     personal

No dependencies found

Pulumi locates its logs in /var/folders/kn/ry_0dbwd215fpwd2rlnxh57c0000gn/T/ by default
warning: Failed to get information about the plugin: failed to discover plugin requirements: unexpected output from 'dotnet list package --include-transitive'. Program does not appear to reference any 'Pulumi.*' packages

```

Frassle:
[...]

PR #688, opened 2025-08-20
Add a log line to debug list package issues

Frassle:
To help with debugging https://github.com/pulumi/pulumi-dotnet/issues/686 and any other similar issues that come up.


Frassle:
Yeh I think `-v10 --logflow --logtostderr` should show them all

PR #705, opened 2025-10-01
use macos-latest instead of macos-13

tgummerer:
macos-13 runners will be retired soon, see
https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/.

The brownout is not supposed to happen until november 4th, but for some reason jobs are not being acquired right now anyway.

Note that this also switches the architecture to arm64 instead of x86_64.  I'd expect most MacOS users to be on that now anyway, so that's probably a good thing to test either way.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -173,7 +173,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-13]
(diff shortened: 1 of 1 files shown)


tgummerer:
Ah the macos failures are because GitHub Actions is having issues right now in general and not because of the `macos-13` deprecation.  That said, `macos-13` is still gonna stop working in two months, so we should do this either way.

PR #711, opened 2025-10-07
Mise for CI: format

i-am-tom:
Continuing from #708 for my ops week treat: I'm just going to start removing setup from CI jobs until everything's done by Mise. Hopefully, these implements are small enough to be uncontroversial.


tgummerer (approved):
Curious, what's the perf difference between the two setups?

PR #736, opened 2025-11-04
implement event log via gRPC for dotnet

tgummerer:
Follow up for https://github.com/pulumi/pulumi/pull/20822, for dotnet.

Assisted by: Claude (it's probably a better dotnet programmer than me, which isn't hard)


sdk/Pulumi.Automation/Events/GrpcEventWatcher.cs
```diff
@@ -0,0 +1,121 @@
...
+    {
+        private readonly IHost _host;
+        private readonly TaskCompletionSource<int> _portTcs = new TaskCompletionSource<int>(TaskCreationOptions.RunContinuationsAsynchronously);
+        private readonly CancellationTokenSource _shutdownCts = new CancellationTokenSource();
+
+        public string EventLogPath { get; private set; } = string.Empty;
+
+        public GrpcEventWatcher(Action<EngineEvent> onEvent, CancellationToken cancellationToken)
+        {
+            _host = Host.CreateDefaultBuilder()
```
Frassle: We've got a few places now that repeat this server setup code, feels like we should probably put this in a util class now. I can do that later if you don't want to muck about with C#.

tgummerer: Yeah, makes sense.  I'd rather not do it in this PR, but opened https://github.com/pulumi/pulumi-dotnet/issues/738.

PR #745, opened 2025-11-09
Fix/687 - NativeAoT Support by upgrading to Dotnet 8 (lts) and implementing `System.Text.Json` source generation

ThaDaVos:
These changes should allow one to use the project with NativeAoT as it now uses the latest LTS (Dotnet 8) and includes source generation for `System.Text.Json`.

Only thing I noticed after getting all `System.Text.Json` stuff working while setting `<JsonSerializerIsReflectionEnabledByDefault>false</JsonSerializerIsReflectionEnabledByDefault>` in my project - is that I get an error from the `grpc` package used:
```
Exception has occurred: CLR/Grpc.Core.RpcException
An exception of type 'Grpc.Core.RpcException' occurred in Pulumi.dll but was not handled in user code: 'Status(StatusCode="Unimplemented", Detail="")'
   at Pulumirpc.LanguageRuntime.LanguageRuntimeBase.GetRequiredPackages(GetRequiredPackagesRequest request, ServerCallContext context) in {PROJECT_PATH}\pulumi-dotnet\sdk\Pulumi\obj\Debug\net8.0\pulumi\LanguageGrpc.cs:line 311
   at Grpc.Shared.Server.UnaryServerMethodInvoker`3.Invoke(HttpContext httpContext, ServerCallContext serverCallContext, TRequest request)
```

Which I cannot figure out the actual cause of


ThaDaVos:
It's been months since I made this PR and I haven't heard anything since, has there been any progress?

Frassle:
We still think bumping to net8 is going to cause significant downstream breakage because of how net builds work, so we're still stuck to targeting net6 for now.

We started making a plan to fix this but it hasn't been an internal priority to actually get it finished, and is made more complicated by the fact that we need to coordinate the fix across multiple internal teams and external provider developers.

ThaDaVos:
Can't you do a bump to DotNet 8 in a major version of the (Automation) SDK? This would partially resolve the issue as most of the time major versions are allowed to break backwards compatibility and instead of targeting DotNet 8, go straight for DotNet 10 LTS to also get some other improvements - also DotNet 8 will be end of support in November this year (https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core)

Frassle:
> Can't you do a bump to DotNet 8 in a major version of the (Automation) SDK?

We have historically avoided major version changes because of a want to keep the engine version and sdk versions tied. That _might_ be less important now so we could consider it.

> go straight for DotNet 10 LTS to also get some other improvements - also DotNet 8 will be end of support in November this year

Yes, we're aware of the dotnet 8 LTS coming to an end soon. A major bump direct to 10 as we drop 8 LTS is definitely worth considering.

ThaDaVos:
If there is help wanted with a PR for DotNet 10 (Including a redo of the source generation here for NativeAoT) let me know, I may be able to lend a hand if I have the time

ThaDaVos:
@Frassle can you check the new PR?
https://github.com/pulumi/pulumi-dotnet/pull/1109

PR #829, opened 2026-02-09
Tag the module on release

iwahbe:
[`pulumi-language-dotnet`](https://github.com/pulumi/pulumi-dotnet/tree/main/pulumi-language-dotnet) is now a module that needs to be consumed downstream, so it should be tagged on release.


Frassle:
> consumed downstream

What's using it downstream? Because we should see about uncoupling that.

Frassle (approved):

PR #884, opened 2026-02-26
Install netcoredbg on Windows via curl

julienp:
We were using scoop to install this, but we're seeing a lot of flakes
when setting up scoop itself due to 503 or 403s.

Download the binary release from GitHub instead and add it to our path.


Frassle:
One less thing that can flake 👍

PR #918, opened 2026-03-20
Fix property path codegen for resource options

iwahbe:
The dotnet codegen was missing `AdditionalSecretOutputs` support entirely, and was incorrectly applying C# keyword escaping to property path strings in `IgnoreChanges`, `HideDiffs`, and `ReplaceOnChanges` (e.g. `"value"` became `"@value"`). Property paths are string contents, not C# identifiers, so they should not be escaped.

Fix by extracting the raw path from `ResourceProperty` traversals instead of formatting through `GenScopeTraversalExpression` (which calls `makeValidIdentifier`). Also updates the pulumi submodule and go.mod to pick up pulumi/pulumi#22302 which types `additionalSecretOutputs` as `ResourcePropertyType`.

Unskips: `l2-resource-option-additional-secret-outputs`, `l2-resource-option-ignore-changes`, `l2-resource-option-delete-before-replace`, `l2-resource-option-hide-diffs`.


iwahbe:
Now ready for a human to review.

pulumi-language-dotnet/codegen/gen_program.go
```diff
@@ -1306,15 +1306,24 @@ func (g *generator) genResourceOptions(opts *pcl.ResourceOptions, resourceOption
...
+			// The entries are property paths resolved as ScopeTraversalExpressions referencing
+			// ResourceProperty definitions. We extract the raw path string to avoid C# keyword
+			// escaping (e.g. "value" must not become "@value").
 			if changes, isTuple := value.(*model.TupleConsExpression); isTuple {
 				g.Fgenf(&result, "\n%s%s =", g.Indent, name)
 				g.Fgenf(&result, "\n%s{", g.Indent)
 				g.Indented(func() {
 					for _, v := range changes.Expressions {
+						if traversal, ok := v.(*model.ScopeTraversalExpression); ok {
+							if prop, ok := traversal.Parts[len(traversal.Parts)-1].(*pcl.ResourceProperty); ok {
```
Frassle: is this right for paths like "nested.path", or "array[0]"? We should probably extend the conformance test to check those

PR #962, opened 2026-04-10
Fix l1-output-null

iwahbe:
Use explicit `new object?[]` instead of `new[]` when array elements include null literals, since C# cannot infer the element type from `new[] { null }`.


pulumi-language-dotnet/codegen/gen_program_expressions.go
```diff
@@ -1269,7 +1282,13 @@ func (g *generator) GenTupleConsExpression(w io.Writer, expr *model.TupleConsExp
 			// because list of a union is mapped to InputList<object>
 			// which means new[] will not work because type-inference won't
 			// know the type of the array beforehand
-			g.Fgenf(w, "%s", g.listInitializer)
+			if g.listInitializer == "new[]" && g.hasNoneTypeElement(expr) {
+				// When elements include null, C# cannot infer the array element type
+				// from `new[] { null }`, so we must use an explicit type.
+				g.Fgen(w, "new object?[]")
```
Frassle, line 1288: Don't we need the list type? `object?` won't always match the destination type will it

Frassle (approved):

Frassle:
This at least unblocks a test, we can add more checks later

iwahbe:
> This at least unblocks a test, we can add more checks later

I'll see how far Claude can get. If it makes real progress, we can review that. Otherwise we can merge as is.

iwahbe:
Claude hasn't made much progress. We can revisit if a user hits this or if it we add a more specific test.

PR #971, opened 2026-04-20
add --run-progam to up in autoapi

jkerken:
Add missing --run-progam to up in automation api


Frassle (approved):
lgtm

Frassle:
xml docs need re-generating and commit. A `dotnet build` should re-gen it.

iwahbe:
We are switching to generating the Automation API, starting with https://github.com/pulumi/pulumi-dotnet/pull/1036. That will resolve this problem.

PR #981, opened 2026-04-29
Fix l1-output-null test

Frassle:
Adds `output?` to array initialisers if there are any nulls.


iwahbe (approved):
Huh, Java had the same bug.

iwahbe:
Isn't this the same as https://github.com/pulumi/pulumi-dotnet/pull/962? Does https://github.com/pulumi/pulumi-dotnet/pull/962#discussion_r3116284776 not apply equally?

Frassle:
Yes :( too much clauding trying to get through these skipped test

PR #983, opened 2026-05-04
upgrade opentelemetry package

### LINKED ISSUE #985: OpenTelemetry.Api 1.9.0 dependency flagged for CVE-2026-40894
The `Pulumi` NuGet package (3.105.0) transitively pulls in `OpenTelemetry.Api` 1.9.0 via the `OpenTelemetry` 1.9.0 reference in `sdk/Pulumi/Pulumi.csproj`, which is flagged by [CVE-2026-40894](https://nvd.nist.gov/vuln/detail/CVE-2026-40894) / [GHSA-g94r-2vxg-569j](https://github.com/open-telemetry/opentelemetry-dotnet/security/advisories/GHSA-g94r-2vxg-569j) (CVSS 5.3, fixed in 1.15.3). We should update this dependency for security reasons.

tgummerer:
The version we're currently using is flagged for a security vulnerability.  Upgrade the package to resolve that.

Fixes https://github.com/pulumi/pulumi-dotnet/issues/985


iwahbe (approved):

sdk/Pulumi.Tests/Pulumi.Tests.csproj
```diff
@@ -17,6 +17,30 @@
     <PackageReference Include="xunit.runner.visualstudio" Version="2.8.2" />
   </ItemGroup>
 
+  <!--
```
Frassle: This feels worrying. Is every customer on net8 going to need to do the same?

tgummerer: Yeah so this was something claude tried, but it didn't work out anyway.  I wanted to try to avoid bumping our target framework to net8.0, but I think it will be necessary, but then we don't need this change.

tgummerer:
I think this needs a re-review by someone that knows dotnet a bit better than me.  I *think* we actually need to bump the target framework to `net8.0`, to avoid a dependency mess, at least claude couldn't figure out a better solution (and I'm not smarter than it with dotnet, that's for sure).  

We'll need to make sure there's not going to be too much of a fallout for @pulumi/providers before we merge this, but I guess this issue kind of forces our hand in doing the upgrade eventually.

sdk/Pulumi.Automation/Events/GrpcEventWatcher.cs
```diff
@@ -105,7 +105,7 @@ public GrpcEventWatcher(Action<EngineEvent> onEvent, CancellationToken cancellat
                 }
             }, cancellationToken);
 
-            _ = _portTcs.Task.Wait(TimeSpan.FromSeconds(10));
+            _ = _portTcs.Task.Wait(TimeSpan.FromSeconds(10), CancellationToken.None);
```
Frassle: Shouldn't this take cancellationToken?

tgummerer: Good catch, yeah it probably should, thanks!

Frassle (approved):

Frassle:
Also lets add a changelog that we're updating to 8

PR #1000, opened 2026-05-08
Fix l1-builtin-string conformance test

### LINKED ISSUE #995: length() returns UTF-16 code units, not graphemes
The `l1-builtin-string` conformance test added in pulumi/pulumi#22802 (shipping with pulumi/pulumi v3.236.0) expects `length()` on strings to return the number of Unicode grapheme clusters. The .NET codegen emits `s.Length`, which returns UTF-16 code units.

For the test's emoji input `"👾-🕹️"`:

- Expected `lengthResult`: `3` (grapheme clusters: `👾`, `-`, `🕹️`)
- Actual `lengthResult`: `6` (`👾` is 2 UTF-16 units, `-` is 1, `🕹️` is 3 with VS-16)

`splitResult`, `joinResult`, and `interpolateResult` all match. Only `length` is wrong.

The upstream test at `pulumi/pulumi/pkg/v3/testing/pulumi-test-language/tests/l1_builtin_string.go` uses `github.com/rivo/uniseg`'s `GraphemeClusterCount` as the spec, and the comment on the PR explicitly notes that "a lot of languages are going to get 'length' wrong" -- this is the expected gap for .NET.

## Suggested fix

In `pulumi-language-dotnet/codegen/gen_program.go`, change the `length()` translation for string arguments to emit `new System.Globalization.StringInfo(s).LengthInTextElements`. Modern .NET (5+) implements `LengthInTextElements` against UAX #29 grapheme cluster boundaries, which matches `uniseg`'s behavior for the corpus the conformance test exercises (multi-byte Latin, emoji with VS-16, ZWJ family sequences).

`length()` on lists and maps should keep using `.Length` / `.Count` -- the bug is specific to the string overload.

## Repro

```
PULUMI_ACCEPT=1 make test_conformance TEST_FILTER='TestLanguage/l1-builtin-string'
```

The generated `Program.cs` translates `length(aString)` to `aString.Length`. The build succeeds, the runtime assertion fails on the emoji case.

Until the codegen is fixed, the test is skipped via `expectedFailures` in `language_test.go`.

iwahbe:
## Summary

- Translate PCL `length()` on strings to `new System.Globalization.StringInfo(s).LengthInTextElements` so it returns Unicode grapheme clusters, matching PCL semantics. `length()` on lists and maps continues to use `.Length` / `.Count`.
- Remove the `l1-builtin-string` skip and add the generated conformance fixture under `pulumi-language-dotnet/testdata/projects/l1-builtin-string/`.

Fixes #995.

## Test plan

- [x] `make test_conformance TEST_FILTER='TestLanguage/l1-builtin-string'` passes
- [x] `make test_codegen` passes (no golden file regressions)
- [x] `make lint_language_host` and `make format_language_host_check` pass


Frassle (approved):
Nice I'd hoped this would be easy to fix

PR #1008, opened 2026-05-12
Fix proxy-index conformance tests

iwahbe:
## Summary

Two remaining codegen fixes, unlocking the l1-proxy-index and l2-proxy-index conformance tests:

- An object literal passed to `secret(...)` is emitted as an anonymous record rather than a `Dictionary`, so the PascalCased property access emitted for Apply-lambda traversals resolves.
- `isJSONElementRoot` (from #1144) gains two arms: a local variable derived from a JsonElement-typed config (e.g. `a = secret(anyObject)`) recurses into its defining expression, and an Apply-lambda parameter resolves through the program-level variable it is named after. Traversals through those Apply lambdas then emit `.GetProperty(...)`.

The l2-proxy-index snapshots were missing or stale (the project snapshot did not exist; the ref-ref SDK snapshot predated the `innerDataList`/`optionalInner` schema fields); both regenerated with `PULUMI_ACCEPT=1`.


iwahbe:

- #1142
    - #1145
        - #1143
            - #1144
                - #1008 ◀

pulumi-language-dotnet/codegen/gen_program_expressions.go
```diff
@@ -840,6 +840,18 @@ func (g *generator) GenFunctionCallExpression(w io.Writer, expr *model.FunctionC
...
+		// lambdas whose codegen emits property access. Emit an anonymous record
+		// instead: its PascalCased properties line up with traversal codegen.
+		if obj, ok := expr.Args[0].(*model.ObjectConsExpression); ok {
+			if _, hasSchema := g.toSchemaType(obj.Type()); !hasSchema {
+				g.Fgen(w, "Output.CreateSecret(")
+				g.genAnonymousRecord(w, obj)
+				g.Fgen(w, ")")
+				return
+			}
+		}
```
Frassle, line 854: Surely this isn't "secret" specific! Do we need better tracking of if "objects" are dicts or actual C# objects.

iwahbe: I think we do, but I think fixing that is a larger refactor.

PR #1016, opened 2026-05-19
Target net8.0 in generated .csproj files

jkodroff:
## Description

Bumps the generated .NET project target framework from `net6.0` to `net8.0`.

`net6.0` reached end of support in November 2024, so generated SDKs should target a currently-supported LTS release. This updates the codegen `.csproj` template and regenerates the affected golden files.

## Changes

- `codegen/templates.go`: `<TargetFramework>net6.0</TargetFramework>` → `net8.0`
- Regenerated codegen golden files (`testdata/**`) accordingly


jkodroff:
**Note: this comment was written by Claude Code, acting on behalf of @jkodroff.**

## Why the `conformance-tests` checks are red

This PR bumps the generated `.csproj` target framework from `net6.0` to `net8.0` and regenerates the affected codegen + conformance golden files.

The `conformance-tests` jobs show **4 failing tests** that are **not caused by this change**:

- `l2-map-keys-adversarial`
- `l2-resource-hook-after-failure`
- `l2-resource-hook-ignore-errors`

These are **already failing on `main`** (see the conformance runs for #1010 and #1007), and they merged anyway — `conformance-tests` is not a required status check. They stem from new conformance tests introduced in pulumi/pulumi v3.239.0 that the .NET SDK/codegen does not yet support:

- `l2-map-keys-adversarial` — a genuine **codegen bug**: the generated `primitive-7.0.0` SDK fails to compile with `error CS0542: 'Invoke': member names cannot be the same as their enclosing type`.
- `l2-resource-hook-after-failure` / `l2-resource-hook-ignore-errors` — a **resource-hook behavior gap** in the SDK.

These are three unrelated fixes and will be addressed in **separate PRs**; bundling them here would mix a one-line framework bump with a codegen naming fix and an SDK hooks investigation.

## Mergeability

This PR is **mergeable pending review**. The ~53 conformance failures that the `net6.0` → `net8.0` change *did* introduce (stale golden snapshots) have been fixed by regenerating the snapshots. The remaining 4 failures pre-date this PR and do not regress `main`.

Frassle:
> net6.0 reached end of support in November 2024, so generated SDKs should target a currently-supported LTS release. This updates the codegen .csproj template and regenerates the affected golden files.

We need to think about this, bumping frameworks has a lot of knock on effects. We're also looking at maybe needing to multi-target. cc @Zaid-Ajaj who's looking at this.

jkodroff:
Per @Zaid-Ajaj:

> I think it is probably fine only from program-gen perspective, however I think if we do this, we want to also bump the sdk-gen to target net8.0 which is currently not possible unless we introduce a major breaking change in the core sdk
[8:23 AM]I would hold until the pulumi sdk and provider sdks start targeting net8.0 before making this change

Frassle:
Yeh if we scope this to programgen I'm pretty confident that's ok.

iwahbe:
We already target .NET 8 in programgen, and bumping it in SDK gen is another thing entirely.

Issue #403, opened 2024-11-21
TestDebuggerAttachDotnet is flaky


Frassle:
### What happened?

This test keeps timing out but seemingly only on macos-12.

### Example

https://github.com/pulumi/pulumi-dotnet/actions/runs/11956317375/job/33330781813?pr=400

### Output of `pulumi about`

N/A

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

justinvp:
Also hit it on macos-13: https://github.com/pulumi/pulumi-dotnet/actions/runs/12207877427/job/34060079149#step:12:252

There are similar failures on Windows, but not clear if it's related to `TestDebuggerAttachDotnet` or some other flake.

PR #1027, opened 2026-06-05
Re-enable TestPluginDebuggerAttachDotnet on macOS

### LINKED ISSUE #706: Renable `TestPluginDebuggerAttachDotnet` on MacOS
It looks like `netcoredbg` doesn't work either because of the arm architecture, or because of the MacOS version we're starting to use in https://github.com/pulumi/pulumi-dotnet/pull/705.  I'm disabling this test on MacOS for now in that PR, we should look into re-enabling it.

iwahbe:
## Why

`TestPluginDebuggerAttachDotnet` was skipped on macOS (#705). The failure (`0x80131c3c`) was an **architecture mismatch**: GitHub's macOS runners and Apple Silicon dev machines are arm64, but Samsung only ships an `osx-amd64` netcoredbg, and a debugger cannot attach to a process of a different architecture.

## What

- **`.mise.toml`** — install netcoredbg as a managed tool via the `http` backend with per-platform URLs. Apple Silicon pulls a community-built arm64 binary ([Samsung/netcoredbg#174](https://github.com/Samsung/netcoredbg/issues/174) tracks the missing official asset); every other platform uses Samsung's official builds.
- **`.github/workflows/pr.yml`** — drop the hand-rolled `curl` netcoredbg install steps and the macOS `PATH` override. `jdx/mise-action` already runs `mise install`, so netcoredbg now comes from `.mise.toml` on every platform (single source of truth, and avoids the amd64 download shadowing the arm64 build).
- **`integration_tests/`** — add a `languagePluginPath` helper that builds `pulumi-language-dotnet` once via `sync.Once` and returns its directory, instead of assuming the binary already exists on disk. Re-enable the test by removing the darwin skip.

`TestDebuggerAttachDotnet` remains skipped on macOS/Windows — that is a separate flakiness issue (#403), not the architecture problem.

## Test

- `TestPluginDebuggerAttachDotnet` passes locally on arm64 macOS using the `.mise.toml`-provided debugger.
- `make format_integration_tests_check` and `make lint_integration_tests` are clean.

Fixes #706

### DIFF
--- a/integration_tests/integration_util_test.go
+++ b/integration_tests/integration_util_test.go
@@ -12,22 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// The linter doesn't see the uses since the consumers are conditionally compiled tests.
-//
-//nolint:unused,deadcode,varcheck
 package integrationtests
 
 import (
-	"bufio"
 	"bytes"
-	"context"
-	"encoding/json"
 	"fmt"
 	"io/fs"
 	"os"
 	"os/exec"
 	"path/filepath"
-	"regexp"
 	"strings"
 	"sync"
 	"testing"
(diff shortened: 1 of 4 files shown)


tgummerer (approved):
Nice, thanks!

integration_tests/integration_util_test.go
```diff
@@ -351,37 +242,6 @@ func testConstructMethodsErrors(t *testing.T, lang string) {
 	})
 }
 
-func testConstructOutputValues(t *testing.T, lang string, dependencies ...string) {
```
Frassle, line 354: AI seems to have done a general cleanup pass here deleting a load of other stuff?

iwahbe: No, that was me. These are private functions that are never called.

PR #1041, opened 2026-06-15
Pin mise to 2026.6.9 to fix vfox:dotnet install on Windows

i-am-tom:
## Problem

All **Windows** CI jobs that provision dotnet through mise (`integration-tests`, `codegen-tests`) fail at the **"Setup mise and install tools"** step:

```
Processing -File '"...\dotnet-install.ps1"' failed: Illegal characters in path.
mise ERROR Failed to install vfox:dotnet@8.0: ... dotnet binary not found at ...\dotnet.exe
```

macOS and Linux are unaffected.

## Root cause

The jobs install the latest mise binary from `mise.jdx.dev/VERSION`, which is now **`2026.6.10`** (released 2026-06-14). Its changelog includes *"Plugin install hooks now run under sanitized environments"* — and on Windows this regresses `vfox:dotnet`: the plugin's post-install hook runs `dotnet-install.ps1` with a quoted path that PowerShell rejects (`Illegal characters in path`), so `dotnet.exe` is never installed and the job dies during setup. `changie` and every other tool still install fine — only `vfox:dotnet` on Windows is affected.

This was masked on `main` by `mise-action`'s cache (the cache key hashes `.mise.toml`). It surfaces as soon as the cache is invalidated — e.g. any PR that touches `.mise.toml`, or normal cache expiry — so `main` is one cache miss away from the same red.

## Fix

Pin the mise binary to **`2026.6.9`** (the last release before the vfox hook change) via a `PINNED_MISE_VERSION` env var in `pr.yml`, referenced from each `mise-action` step's `version:` input. A single comment documents why and when to remove it.

This is validated automatically: `integration-tests` and `codegen-tests` run a hardcoded `windows-latest` matrix leg, so this PR's own CI exercises the fix.

## Notes for reviewers

- Only `pr.yml` is touched (7 `mise-action` steps). `changelog.yml` / `attempt-release.yml` are Linux-only and unaffected.
- Pure CI change — no SDK/runtime impact, so it likely needs the `impact/no-changelog-required` label rather than a changelog fragment.
- Revert is trivial once the upstream regression is fixed: delete the env var and the `version:` inputs.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -17,16 +17,14 @@ env:
   ESC_ACTION_OIDC_REQUESTED_TOKEN_TYPE: urn:pulumi:token-type:access_token:organization
   ESC_ACTION_ENVIRONMENT: github-secrets/pulumi-pulumi-dotnet
   ESC_ACTION_EXPORT_ENVIRONMENT_VARIABLES: PULUMI_ACCESS_TOKEN=PULUMI_ACCESS_TOKEN_PRODUCTION
-  # Pin the mise binary: recent releases regressed `vfox:dotnet` on Windows.
-  #   * 2026.6.10 fails inside the plugin ("Illegal characters in path" running
-  #     dotnet-install.ps1), before the postinstall hook even runs.
-  #   * 2026.6.9 installs the SDK but then fails the .mise.toml `postinstall`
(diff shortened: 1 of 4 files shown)


iwahbe (approved):
I'm merging to unblock CI.

PR #1049, opened 2026-06-18
Automation API Codegen #2: Generating commands

i-am-tom:
```csharp
/// <summary>
/// Builds the arguments for the <c>pulumi cancel</c> command.
/// </summary>
public IReadOnlyList<string> Cancel(string? stackName = null, PulumiCancelOptions? options = null)
{
    var __final = new List<string> { "cancel" };
    __final.Add("--yes");
    if (options?.Color is { } __valueColor)
    {
        __final.Add("--color");
        __final.Add(__valueColor);
    }
    // ... --stack, --verbose ...
    var __arguments = new List<string>();
    if (stackName != null)
    {
        __arguments.Add(stackName);
    }

    if (__arguments.Count > 0)
    {
        __final.Add("--");
        __final.AddRange(__arguments);
    }

    return __final;
}
```


iwahbe (approved):
Looks good to me.

.NET isn't my specialty, but I think we are good to merge.

PR #1065, opened 2026-07-03
Drop the pulumi/pulumi submodule

iwahbe:
> [!NOTE]
> Stacked on #1076: this branch is rebased onto `iwahbe/error-hook-conformance` and targets it as the PR base. It inherits that PR's pin to an unreleased pulumi/pulumi master commit (`cd588358a1`); the sync scripts below handle pseudo-version pins for that reason.

## Summary

Removes the `pulumi` git submodule. Everything it provided is replaced by an alternative pinned to the same source of truth: the `github.com/pulumi/pulumi/{pkg,sdk}/v3` version in `pulumi-language-dotnet/go.mod`.

The `.proto` files cannot be sourced from the Go module graph — the `pkg` and `sdk` Go modules ship only the generated Go code (`sdk/v3/proto/go`), not the `.proto` sources, which live at the repo root of pulumi/pulumi outside both module boundaries. They are vendored instead.

## Changes

- **Protobuf definitions** (`sdk/Pulumi/Pulumi.csproj`): vendored under `proto/` at the repo root, along with the Automation API CLI-spec overrides file (`tools/automation/automation-overrides.json` → `sdk/Pulumi.Automation.Codegen/`). `make sync_protos` re-syncs both from a sparse clone of pulumi/pulumi at the pinned version. The `format` CI job re-runs the sync and fails on drift, so a `go.mod` bump without a proto sync cannot merge.
- **`make generate_cli_spec`**: builds the pulumi CLI through a Go `tool` directive in `pulumi-language-dotnet/go.mod` (`go tool pulumi generate-cli-spec`). `go run` of the dependency's main package fails under module graph pruning (the CLI's dep closure isn't in `go.sum`), and `go run pkg@version` is rejected because `pkg/v3`'s go.mod contains `replace` directives. The tool directive records the full closure and moves in lockstep with the `pkg/v3` requirement.
- **Codegen test corpus**: the codegen tests read the shared corpus (schemas, PCL programs, `types.json`) from `tests/testdata/codegen`. At 11MB across ~2,250 files it is too large to vendor, so `make test_codegen` fetches it into the gitignored `pulumi-language-dotnet/codegen/testdata/upstream/` via a sparse clone of the pinned version, with a version stamp to skip re-downloads.
- **Pseudo-version pins** (`scripts/checkout_pulumi.sh`): the sparse checkout used by `sync_protos.sh` and `get_codegen_testdata.sh` is factored into a shared helper. A pseudo-version pin (like the one inherited from #1076) has no tag to clone, so the helper resolves it to its commit through the Go module proxy and fetches by hash.
- Drop the submodule (`.gitmodules` + gitlink) and all `submodules: recursive` checkouts from CI.
- Update README/AGENTS.md references.

An earlier revision vendored `pulumi/tests/testprovider` as `integration_tests/testprovider_go` for `TestDotnetErrorHooks`; that test was replaced by the `l2-resource-hook-on-error` conformance test in #1076, so the vendored provider is gone again.

## Testing

- `make build_sdk` compiles against the vendored protos (no `Pulumi.xml` drift).
- `make test_codegen`: 235 tests pass, 18 skipped, with the corpus fetched at the pseudo-version.
- `make test_automation_codegen`: 57 tests pass (spec generated via `go tool pulumi`).
- `make test_conformance TEST_FILTER=TestLanguage/l2-resource-hook-on-error` passes.
- `make lint_language_host`, `make lint_integration_tests` clean.
- `scripts/sync_protos.sh` round-trips with no diff at the pinned commit.


Frassle (commented):
I'm really not convinced this is better than just the submodule.
The testdata changes are fine, if we want to pull that via go instead of the submodule that makes sense. But for the proto files the submodule seems way simpler.

iwahbe:
I'm not happy with a bunch of stuff in this PR, and I haven't read through carefully. Package gen tests are a real drag here.

PR #1066, opened 2026-07-07
Automation API Codegen #3: Boilerplate

i-am-tom:
See also:
* https://github.com/pulumi/pulumi/pull/22059
* https://github.com/pulumi/pulumi/pull/22490

```csharp
// generated (Commands.cs)
public Task<CommandResult> Cancel(
    string? stackName = null,
    PulumiCancelOptions? options = null,
    CancellationToken cancellationToken = default)
{
    var __final = new List<string> { "cancel" };
    __final.Add("--yes");
    // ... build args ...
    return RunAsync(__final, options, cancellationToken);   // <- provided by the boilerplate
}
```


iwahbe (approved):
It looks reasonable and LLMs like it.

I'm not a C# expert but I'm approving to unblock.

PR #1068, opened 2026-07-07
Add extension codegen

guineveresaenger:
This pull request adds codegen capability for provider parameterization extensions.

Part of https://github.com/pulumi/pulumi/issues/17059.


pulumi-language-dotnet/codegen/testdata/schemas/extension-parameterized-resource/schema.json
Frassle, line 1: We shouldn't need to do a codegen test for this, surely we can just pull the new conformance test and have this covered by that? We really want to move away from most "golden file codegen" tests, they don't really check enough to be worth it.

iwahbe (approved):

PR #1076, opened 2026-07-17
Generate resource hooks in dotnet programs; adopt the error hook conformance test

iwahbe:
Implements resource hook code generation for .NET programs and adopts the `l2-resource-hook-on-error` conformance test added in pulumi/pulumi#23839.

This PR removes the now redundant `TestDotnetErrorHooks`, the conformance test makes it redundant.


pulumi-language-dotnet/testdata/projects/l2-resource-hook-on-error/Program.cs
```diff
@@ -0,0 +1,37 @@
...
+        try
+        {
+            var process = Process.Start("touch", new[] { hookTestFile });
+            process.WaitForExit();
+            return Task.FromResult(process.ExitCode == 0);
+        }
+        catch
+        {
+            return Task.FromResult(false);
+        }
```
Frassle: Should this catch? Shouldn't this just let the exception throw so the hook call is marked as failed?

iwahbe: No, it should not. Now fixed.

Frassle (approved):

Issue #1102, opened 2026-08-04
.NET SDK strips `__`-prefixed keys from invoke return values


EmmittJ:
### What happened?

I was creating a provider resource whose input is a descriptor object keyed by a `__`-prefixed type
discriminator, using the helper invoke the provider ships for building that descriptor. The
documented pattern is to pass the invoke's result directly into the resource, as the TypeScript and
YAML examples do.

In C# this fails:

```
error: property permissions value {map[permissions:{[...]}]} has a problem:
       failed to parse permission descriptor: type '' not recognized
```

The behaviour originates in the .NET SDK rather than the provider.
`Deserializer.DeserializeStruct` drops every `__`-prefixed key when materializing invoke return
values, so the discriminator is removed before the result reaches user code. Passing that value into
the resource then fails, because the shape no longer matches what the provider expects.

**Expected:** invoke results are deserialized as the provider's schema defines them, including any
`__`-prefixed fields.

**Actual:** `__`-prefixed fields are removed, without an error or warning. The failure surfaces
later, reported by the provider and referencing the provider and property names.

This appears to be the .NET counterpart of pulumi/pulumi#22738 ("Python SDK strips `__`-prefixed
keys from invoke return values"), which was accepted and fixed. That issue states that
"TS/Go/.NET/Java/YAML do not strip these keys"; the repro below shows .NET does strip them. The .NET
implementation also differs from Python's pre-fix behaviour in two ways: there is no
`keep_internal`-style parameter, and no `__provider` exemption.

Resources involved: any provider invoke whose result schema uses a `__`-prefixed field. I
encountered it on `pulumiservice.OrganizationRole` via its `BuildAllowPermissions` invoke, though
nothing about the behaviour is specific to that provider.

### Example

Self-contained: no provider plugin, no cloud account, no credentials, no stack. The only package
reference is `Pulumi`, and `IMocks` supplies the invoke result directly.

`repro.csproj`:

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net10.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Pulumi" Version="3.110.0" />
  </ItemGroup>

</Project>
```

`Program.cs`:

```csharp
using System.Collections.Immutable;
[...]

PR #1107, opened 2026-08-06
Dispatch discriminated union deserialization on the discriminator tag

### LINKED ISSUE #1104: Union output deserialization selects the first structurally convertible member instead of dispatching on the discriminator
`Converter.TryConvertOneOf` (sdk/Pulumi/Serialization/Converter.cs) converts against T0 and returns on the first success. Missing keys default silently in `TryConvertObject`, so for object unions the first member converts even when the discriminator names another member.

This is masked today because 3+-member unions type as `object` (#1098). Once #1098 lands, union outputs will deserialize as the wrong variant; the `l2-discriminated-union-many` same-shape-variants conformance case covers exactly this.

Fix: dispatch on the schema discriminator tag. Should land with #1098.

lukebward:
Fixes #1104.

`Converter.TryConvertOneOf` converts against `T0` and returns on the first success; `TryConvertObject` tolerates missing keys, so for object unions the first member usually wins even when the discriminator names another. Two variants that differ only by tag cannot deserialize correctly today.

This adds a discriminator-driven path:

- `DiscriminatedUnionDiscriminatorAttribute(string propertyName)` and `DiscriminatedUnionCaseAttribute(string tag, Type type)` on a generated interface; concrete `[OutputType]` classes implement the interface.
- `Converter.TryConvertObject` resolves interface targets carrying the attribute by reading the discriminator property and delegating to the matching case type. `CheckTargetType` validates each case recursively and requires assignability to the interface.
- The same dispatch is added to `PropertyValueSerializer.DeserializeValue` for the component-provider path.
- `Union<T0,T1>` semantics are unchanged.

Unknown and missing tags produce errors naming the property and listing the valid tags.

Codegen that emits these attributes is stacked in the follow-up PR; this PR only adds runtime support, so it is inert until an SDK release carries the attributes.

## Test plan

New `DiscriminatedUnionConverterTests` covers tag dispatch where two cases have identical property shapes and differ only by tag (the case first-match cannot solve), all tags round-tripping, unknown-tag and missing-discriminator errors, nesting inside an `[OutputType]`, and secret-wrapped values. `PropertyValueTests` gains the same-shape and unknown-tag cases for the serializer path.

`dotnet test` for `Pulumi.Tests`: 413 passed, 0 failed. `dotnet format --verify-no-changes` clean.

Note for reviewers: the `Converter` fixtures use a plain discriminator name rather than a `__`-prefixed one because `Deserializer.DeserializeStruct` skips keys beginning with `__`, so such a key can never reach the converter on the output path (related: #1102).

### DIFF
--- a/sdk/Pulumi.Tests/Core/PropertyValueTests.cs
+++ b/sdk/Pulumi.Tests/Core/PropertyValueTests.cs
@@ -894,4 +894,60 @@ public class PropertyValueTests
         var tags = await result.Tags.ToOutput().DataTask;
         Assert.False(tags.IsKnown, "tags should be unknown");
     }
+
+    // "password" and "token" share an identical property shape; only the discriminator
+    // can tell them apart.
+    [DiscriminatedUnionType("__type")]
+    [DiscriminatedUnionCase("password", typeof(PasswordCredential))]
+    [DiscriminatedUnionCase("token", typeof(TokenCredential))]
+    [DiscriminatedUnionCase("certificate", typeof(CertificateCredential))]
+    public interface ICredential
+    {
+    }
+
+    public class PasswordCredential : ICredential
+    {
+        public string Data { get; set; } = string.Empty;
+    }
+
+    public class TokenCredential : ICredential
+    {
+        public string Data { get; set; } = string.Empty;
+    }
+
+    public class CertificateCredential : ICredential
+    {
+        public string Data { get; set; } = string.Empty;
+        public string? Thumbprint { get; set; }
+    }
+
+    [Fact]
+    public async Task DeserializingDiscriminatedUnionDispatchesOnTag()
+    {
+        var serializer = CreateSerializer();
+        var data = Object(
+            Pair("__type", new PropertyValue("token")),
+            Pair("Data", new PropertyValue("abc")));
+
+        var credential = await serializer.Deserialize<ICredential>(data);
+
+        var token = Assert.IsType<TokenCredential>(credential);
+        Assert.Equal("abc", token.Data);
+    }
+
+    [Fact]
+    public async Task DeserializingDiscriminatedUnionWithUnknownTagThrows()
+    {
+        var serializer = CreateSerializer();
+        var data = Object(
+            Pair("__type", new PropertyValue("oauth")),
+            Pair("Data", new PropertyValue("abc")));
+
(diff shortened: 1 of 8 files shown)


sdk/Pulumi.Tests/Core/PropertyValueTests.cs
```diff
@@ -894,4 +894,60 @@ public async Task DeserializingUnknownOutputInputMapWorks()
         var tags = await result.Tags.ToOutput().DataTask;
         Assert.False(tags.IsKnown, "tags should be unknown");
     }
+
+    // "password" and "token" share an identical property shape; only the discriminator
+    // can tell them apart.
+    [DiscriminatedUnionType("__type")]
```
Frassle:
```suggestion
    [DiscriminatedUnionType("type")]
```

sdk/Pulumi/Serialization/Attributes.cs
```diff
@@ -110,6 +111,44 @@ public sealed class EnumTypeAttribute : Attribute
...
 
+    /// <summary>
+    /// Attribute used by a Pulumi Cloud Provider Package to mark an interface that represents a
+    /// discriminated union of complex output property types. The interface must also carry one
+    /// <see cref="DiscriminatedUnionCaseAttribute"/> per union member. When deserializing, the
+    /// Pulumi runtime reads the discriminator property named by <see cref="PropertyName"/> from
+    /// the raw value and instantiates the member whose tag matches.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Interface)]
+    public sealed class DiscriminatedUnionTypeAttribute : Attribute
```
Frassle:
```suggestion
    public sealed class DiscriminatedUnionDiscriminatorAttribute : Attribute
```
Matches what we call this in schema

sdk/Pulumi/Serialization/Attributes.cs
```diff
@@ -110,6 +111,44 @@ public sealed class EnumTypeAttribute : Attribute
     {
     }
 
+    /// <summary>
+    /// Attribute used by a Pulumi Cloud Provider Package to mark an interface that represents a
```
Frassle: Not cloud provider specific

Frassle (approved):

PR #1112, opened 2026-08-10
Have Renovate write changelog fragments for its updates

i-am-tom:
After some bashing, I think I'm happy with this - we use mostly Renovate's own templating system to generate the CHANGELOGs. They're not the most verbose, but I think that's fine.


Frassle:
> This uses a GitHub Action rather than a Renovate postUpgradeTasks hook (the pulumi/pulumi approach) because our changie format needs the PR number in the fragment filename and entry link, and Renovate doesn't know it when it creates the branch. It also avoids touching the self-hosted bot's allowed-commands config.

Can we fix that. The pu/pu ones don't need committing with the PR number, if we fix dotnet ones to also not need that (using the same thing we do in pu/pu) then we can changelogs here the same was as pu/pu as well.

Consistency is nice.

Frassle (approved):

PR #1116, opened 2026-08-11
Fix program codegen for conversions between primitive types

julienp:
While trying to update pulumi-dotnet in pulumi/pulumi, a smoke test failed https://github.com/pulumi/pulumi/pull/24249

```
Error:
D:\a\_temp\test-env3863622462\out\Rc.cs(29,25): error CS0029: Cannot implicitly convert type 'Pulumi.Output<string>' to 'Pulumi.Output<dynamic>' [D:\a\_temp\test-env3863622462\out\test-env3863622462.csproj]
```

We need to properly handle the IDType in codegen that was introduced in 3.256.0.

Getting `l2-id-type` passing required dealing with primitive conversions, so that solved a couple other tests along with it.

Int to number conversions now always emit a double literal (`1.0`) when it's a literal, or a `(double)` cast otherwise, because C#'s implicitly-typed arrays infer int[] from bare int elements, but that doesn't convert to `InputList<double>`. Codegen can't tell scalar from array-element context, so this is always applied, not just in arrays.

I also ran into the issue that we can't have a class called `Invoke` with a method named `Invoke`. That's now disambiguate with a `Function` suffix (they are static methods on a static class).


pulumi-language-dotnet/testdata/projects/l2-id-type/Program.cs
```diff
@@ -0,0 +1,87 @@
...
+using Primitive = Pulumi.Primitive;
+
+return await Deployment.RunAsync(() => 
+{
+    // Test that the ID type is treated the same as a string type, despite being tracked as a distinct type. This 
+    // includes directly passing it to string fields, but also for bool and numeric values being able to cast to it.
+    var source1 = new Primitive.Resource("source1", new()
+    {
+        Boolean = false,
+        Float = (double)1,
```
Frassle: This is a bit sad, can we output 1.0 instead of putting double casts.

julienp: Oh good idea, that works nicely for literals

Frassle (approved):

PR #1119, opened 2026-08-11
Avoid allocations when normalizing common Pulumi tokens

corymhall:
.NET program generation normalizes package tokens while building schema lookup maps. Most tokens do not use the default `index` module, but `canonicalToken` currently splits and rebuilds every valid token.

This change returns the original string for the common case and allocates a new string only when it must convert `pkg:index:member` to `pkg::member`.

Measured across the complete AWS PCL corpus of 3,736 examples:

| Metric | Before | After | Change |
|---|---:|---:|---:|
| .NET generation | 15,555 ms | 8,584 ms | -44.8% |
| User CPU | 20,157 ms | 10,005 ms | -50.4% |
| System CPU | 312 ms | 141 ms | -54.8% |
| Allocated bytes | 19.97 GB | 6.30 GB | -68.5% |
| Allocations | 178,920,998 | 19,316,973 | -89.2% |
| GC cycles | 120 | 52 | -56.7% |
| GC pause | 10.064 ms | 3.726 ms | -63.0% |
| Peak RSS | 1.00 GB | 2.14 GB | +113.1% |
| Parse time | 2,828 ms | 2,752 ms | -2.7% |
| Bind time | 8,975 ms | 8,804 ms | -1.9% |

Peak RSS varied substantially between complete benchmark runs and increased in this individual run. The generated files, diagnostics, and workload counts remained unchanged.

Validation:

- `mise exec -- make format_language_host_check`
- `mise exec -- make lint_language_host`
- `mise exec -- make test_codegen`

### DIFF
--- a/pulumi-language-dotnet/codegen/gen_program.go
+++ b/pulumi-language-dotnet/codegen/gen_program.go
@@ -117,14 +117,19 @@ func (g *generator) packageForToken(token, fallback string) string {
 // "index" module and an elided module compare equal — schema tokens spell the
 // module "index" while PCL tokens omit it (e.g. "pkg::member").
 func canonicalToken(token string) string {
-	parts := strings.SplitN(token, ":", 3)
-	if len(parts) != 3 {
+	packageEnd := strings.IndexByte(token, ':')
+	if packageEnd == -1 {
 		return token
 	}
-	if parts[1] == "index" {
-		parts[1] = ""
+	moduleEnd := strings.IndexByte(token[packageEnd+1:], ':')
+	if moduleEnd == -1 {
+		return token
 	}
-	return parts[0] + ":" + parts[1] + ":" + parts[2]
+	moduleEnd += packageEnd + 1
+	if token[packageEnd+1:moduleEnd] != "index" {
+		return token
+	}
+	return token[:packageEnd+1] + token[moduleEnd:]
 }
 
 func (g *generator) resetListInitializer() {
--- a/pulumi-language-dotnet/codegen/gen_program_test.go
+++ b/pulumi-language-dotnet/codegen/gen_program_test.go
@@ -105,6 +105,7 @@ func TestCanonicalToken(t *testing.T) {
 		{token: "aws:index:Bucket", want: "aws::Bucket"},
 		{token: "aws::Bucket", want: "aws::Bucket"},
 		{token: "malformed", want: "malformed"},
+		{token: "aws:index", want: "aws:index"},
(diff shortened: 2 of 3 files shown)


corymhall:

- #1119 ◀
    - #1120

pulumi-language-dotnet/codegen/gen_program_test.go
```diff
@@ -94,6 +94,28 @@ func TestGenerateProgramYAML(t *testing.T) {
...
+
+	tests := []struct {
+		token string
+		want  string
+	}{
+		{token: "aws:s3/bucket:Bucket", want: "aws:s3/bucket:Bucket"},
+		{token: "aws:index:Bucket", want: "aws::Bucket"},
+		{token: "aws::Bucket", want: "aws::Bucket"},
+		{token: "malformed", want: "malformed"},
+		{token: "aws:index:Bucket:extra", want: "aws::Bucket:extra"},
```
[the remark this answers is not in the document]
corymhall: Added the single-colon case (`aws:index`) to cover this early-return branch. The focused test and format check pass.

pulumi-language-dotnet/codegen/gen_program.go
```diff
@@ -117,14 +117,19 @@ func (g *generator) packageForToken(token, fallback string) string {
...
 		return token
 	}
-	if parts[1] == "index" {
-		parts[1] = ""
+	moduleEnd := strings.IndexByte(token[packageEnd+1:], ':')
+	if moduleEnd == -1 {
+		return token
+	}
+	moduleEnd += packageEnd + 1
+	if token[packageEnd+1:moduleEnd] != "index" {
```
Frassle, line 129: If you want to be really quick
```
ref char c = ref Unsafe.Add(ref MemoryMarshal.GetReference(token.AsSpan()), packageEnd+1);
bool isIndex = MemoryMarshal.CreateReadOnlySpan(ref c, moduleEnd - packageEnd - 1).SequenceEqual("index");
```

pose: @Frassle is that meant to be the generated code or the snippet we should try to replicate in Go? Thanks! 🙏

Frassle: Oh I was stuck in C# mode 😆 forgot this was Go code, ignore me I was trying to save string slice allocations but that's less a thing in Go.

iwahbe (approved):
Might be worth adding a `schema.Token` type in pulumi/pulumi to handle this, so canonicalization has a single source of truth.

In the mean time, this looks good.

PR #1142, opened 2026-08-26
Fix codegen for empty array literals

iwahbe:
An empty `new[] {}` is a compile error in every context (CS0826: the
element type can never be inferred). Emit `new object?[] {}` when the
default array initializer is active; target-typed initializers are
unaffected. This unlocks the l1-output-array conformance test.


iwahbe:

- #1142 ◀
    - #1145
        - #1143
            - #1144
                - #1008

pulumi-language-dotnet/codegen/testdata/invoke-inside-conditional-range-pp/invoke-inside-conditional-range.cs
```diff
@@ -9,9 +9,9 @@
 {
     var config = new Config();
     // A list of availability zones names or ids in the region
-    var azs = config.GetObject<string[]>("azs") ?? new[] {};
+    var azs = config.GetObject<string[]>("azs") ?? new object?[] {};
```
iwahbe, line 12: This is still wrong, but its less wrong then before.

Frassle (approved):

PR #1143, opened 2026-08-26
Fix codegen for object literals in typeless positions

iwahbe:
## Summary

A schema-less object literal in a typeless position — an output value, a local variable, or a config default — was emitted as a bare `{ {k, v} }` collection initializer (or `null` when empty), which only compiles against a target-typed collection property.

Those sites now mark such literals with an `__untypedObjectLiteral` intrinsic during lowering (recursing through conversion wrappers and list elements), and rendering emits them as a `Dictionary<string, object?>` that carries its own type. Target-typed positions — resource args and options, component args, invoke args — are never marked and keep the bare initializer, so their generated code is byte-identical.

This unlocks the l1-output-map conformance test.

## Testing

- `TestLanguage/l1-output-map` passes (compile + runtime outputs).
- The four snapshot-sensitive passing tests (l2-plain-component, l2-resource-option-env-var-mappings, l2-resource-primitive-conversions, l2-resource-invoke-dynamic-function) regenerate byte-identical.
- `TestGenerateProgram` golden tests pass with no drift.


iwahbe:

- #1142
    - #1145
        - #1143 ◀
            - #1144
                - #1008

Frassle (approved):

PR #1144, opened 2026-08-26
Fix codegen for typed and untyped config variable traversals

iwahbe:
## Summary

Three fixes for config variables, unlocking the l1-config-types-object conformance test:

- Auto-generated config classes used the wire-format property name directly, mismatching the PascalCased traversal codegen. The C# property is now PascalCased and tagged with `[JsonPropertyName(...)]` so JSON deserialization still matches the on-wire key.
- `config.RequireObject<dynamic>(...)` returns a `JsonElement` boxed as `dynamic`, after which dynamic member access throws at runtime. Config `any` now binds to `JsonElement`, traversals emit `.GetProperty(...)`, numeric conversions emit `.GetDouble()`, and defaulted `any` config binds `JsonElement?` with a `JsonSerializer.SerializeToElement(...)` default.
- Traversals into map-typed values emitted PascalCased property access; a `MapType` source is a Dictionary, so indexer syntax is emitted.

Whether a traversal root is a JsonElement is derived from the model, not tracked in generator state: traversal expressions carry their referent (`Parts[0]`), so `isJSONElementRoot` is a pure predicate on the referenced config variable's declared type. #1008 extends the same predicate to locals and Apply-lambda parameters.

## Testing

- `TestLanguage/l1-config-types-object` passes (compile + runtime outputs).
- `TestGenerateProgram` golden tests pass with no drift.


iwahbe:

- #1142
    - #1145
        - #1143
            - #1144 ◀
                - #1008

Frassle:
The first and last fixes seem sensible here. But changing from dynamic to JsonElement feels off. Might be better to switch to `System.Dynamic.ExpandoObject` with a cast back to dynamic, that should then behave as expected for the rest of the program.

PR #1145, opened 2026-08-26
Fix codegen for empty list and map literals in resource args

iwahbe:
An empty list in a resource argument emitted `new[] {}`, which cannot
infer its element type, and an empty map literal emitted `null`, which
fails at runtime for required properties (`{}` is an empty map value,
not null). Resource argument properties are typed, so emit target-typed
`new()` / `new() { }` for both, scoped to the resource input loop.

This unlocks the l2-resource-provider-inheritance conformance test.


iwahbe:

- #1142
    - #1145 ◀
        - #1143
            - #1144
                - #1008

Frassle (approved):

PR #1147, opened 2026-08-27
Generalize representation tracking for schema-less object literals

iwahbe:
Follow-up to the discussion in https://github.com/pulumi/pulumi-dotnet/pull/1008#discussion_r3871577763: the dictionary-vs-record mismatch is not `secret`-specific, so this generalizes the tracking.

Producer sites and the traversal codegen previously decided the C# representation of a schema-less object literal independently, one mechanism per site: local variables recorded their typed-Dictionary choice in the generator-side `typedDictionaryLocals` map, and the `secret(...)` case embedded its own anonymous-record check, while traversal spelling relied on those decisions matching by construction.

Representation decisions now work in two coordinated halves:

- **One classifier.** `exprContainerRepr` maps a defining expression to its container representation (Dictionary or anonymous record), and `rootContainerRepr` resolves a traversal root through its definition to the same answer, so the container shape and the accessor spelling cannot diverge. The `typedDictionaryLocals` map is gone.
- **Decisions ride the tree as lowering-time marks.** Following the pattern `__untypedObjectLiteral` established, a rewrite during lowering wraps the schema-less object literal argument of each `secret(...)` call in a new `__anonymousRecordLiteral` intrinsic, so the `secret` emission is just `Output.CreateSecret(%v)` and any future Output-wrapping builtin gets record rendering by marking alone. The `__untypedObjectLiteral` intrinsic now carries the Dictionary value type, which lets `genLocalVariable` express its typed-Dictionary choice as a mark and deletes `genTypedLocalObjectConsExpression`.

The JsonElement tracking (`isJSONElementRoot` and friends) is deliberately left as a separate axis: whether Dynamic-typed hops hold a `JsonElement` is orthogonal to the container shape and can be true at the same time as either representation (e.g. `secret({a = anyObject.x})`), so folding it into the same enum would lose information.

The generalization also fixes a latent bug: a schema-less object literal nested inside `secret({...})` previously emitted a bare collection initializer, which neither compiled nor matched the property-access spelling of its traversals. `genAnonymousRecord` now recurses, and a new unit test asserts the full generated program for that case (verified to fail on the base branch).

No golden files change: all codegen golden tests and the conformance snapshots for the affected tests (`l1-proxy-index`, `l2-proxy-index`, `l1-output-map`, `l1-config-types-object`, `l1-builtin-list`, `l2-id-type`, `l1-output-secret`) are untouched.

### DIFF
--- a/.changes/unreleased/bug-fixes-1147.yaml
+++ b/.changes/unreleased/bug-fixes-1147.yaml
@@ -1,0 +1,6 @@
+component: runtime
+kind: bug-fixes
+body: Fix program codegen for schema-less object literals nested inside `secret(...)`
+time: 2026-08-27T12:00:00.000000+00:00
+custom:
+    PR: "1147"
--- a/pulumi-language-dotnet/codegen/gen_intrinsics.go
+++ b/pulumi-language-dotnet/codegen/gen_intrinsics.go
@@ -16,6 +16,7 @@ package dotnet
 
 import (
 	"github.com/pulumi/pulumi/pkg/v3/codegen/hcl2/model"
+	"github.com/zclconf/go-cty/cty"
 )
 
 const (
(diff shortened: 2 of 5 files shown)


iwahbe:

- #1142
    - #1145
        - #1143
            - #1144
                - #1008
                    - #1147 ◀

PR #1154, opened 2026-09-03
Generate code for `read` PCL blocks

Frassle:
The .NET program generator dispatched only on `*pcl.Resource` in `genNode`, so `read "..." "..." { ... }` blocks (which the PCL binder represents as `*pcl.ReadResource`) were silently dropped from the emitted C# program. Downstream references to the read resource's variable then failed to compile with CS0103.

Add a `genReadResource` that emits `Type.Get(name, id, state, options)` using the static factory that generated .NET resources already expose, splitting the `id` attribute from the remaining inputs to construct `TypeState`. Wire it into `genNode`, `usingStatements`, and the component-nested node loop, and refactor `resourceTypeName` / `resourceArgsTypeName` to share a token-based helper reusable for both node kinds.

Unblocks the l2-resource-read conformance test.


pulumi-language-dotnet/language_test.go
iwahbe, line 213: Does this test also unlock?

iwahbe (approved):

pulumi-language-dotnet/language_test.go
Frassle: No, that needs some sdk work

PR #1166, opened 2026-09-09
unpin .net 10.0

tgummerer:
We pinned this previously to fix issues in integration tests.  Unpin it to see if a new version of .net fixed the issues.


julienp (approved):
looks like it's working!

PR #1173, opened 2026-09-14
Switch to Google.Protobuf

Frassle:
New versions of Google.Protobuf include options to change the recursion limit. We can use this instead of our outdated fork. This should also simplify dependency management as we're no longer trying to fit our library into the dependency tree as a version of Google.Protobuf which could get nuget confused in some cases.


sdk/Pulumi.Tests/Serialization/ProtobufRecusionLimitTests.cs
```diff
@@ -13,16 +14,16 @@ public void InvokeResponseRecursionLimitTest()
         {
             var response = new Pulumirpc.InvokeResponse
             {
-                Return = GenerateNestedStruct(50),
+                Return = GenerateNestedStruct(300),
             };
-            var serialized = response.ToByteArray();
+            var serialized = response.ToByteString();
```
iwahbe, line 19: Why are we changing the type here?

sdk/Pulumi/Pulumi.xml
```diff
@@ -2016,6 +2016,15 @@
             in outgoing call metadata so the engine can parent its spans correctly.
             </summary>
         </member>
+        <member name="T:Pulumi.ProtobufRecursionLimitInterceptor">
```
iwahbe, line 2019: I'm sure I'm missing something about C#, but can we make this private?

iwahbe (approved):

sdk/Pulumi.Tests/Serialization/ProtobufRecusionLimitTests.cs
Frassle: Because everywhere we call the new Parse method in the real code we have a ByteString not a []byte. So it made sense to just support ByteString and change this test to match.

sdk/Pulumi/Pulumi.xml
Frassle: It is, it's an internal class but those still get added to the docs.

